PR #2, opened 2020-09-04
Fixed _run_flyers() for sirepo optimization

AbbyGi:


bloptools/de_opt_utils.py
```diff
@@ -123,9 +123,16 @@ def calc_velocity(motors, dists, velocity_limits, max_velocity=None, min_velocit
...
+    if isinstance(flyers[0], list):
+        for flyer in flyers:
+            # pass flyer into bp.fly since flyer is already a list of flyers
+            uids = (yield from bp.fly(flyer))
+            uid_list.append(uids)
+    else:
+        for flyer in flyers:
+            # pass [flyer] into bp.fly since flyer isn't a list here
+            uid = (yield from bp.fly([flyer]))
+            uid_list.append(uid)
```
mrakitin, line 135: Instead of repeating the code, can we redefine `flyers = flyers[0]` in this case? Also, we should check that the flyers list is not empty (which is potentially possible input, but makes no sense for simulations).

mrakitin (commented):
I have a [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) suggestion below. Otherwise, it's a great catch!

mrakitin (approved):

mrakitin:
Thanks, @AbbyGi! That's a good workaround for now.

PR #7, opened 2023-03-02
Update CI and the skeleton using scientific-python-cookiecutter PR105

mrakitin:
https://github.com/NSLS-II/scientific-python-cookiecutter/pull/105.


thomaswmorris (commented):
all good!

PR #8, opened 2023-03-06
Fisher matrix

thomaswmorris:
This PR does a few things. It's called "fisher-matrix" because it adds the ability to do optimal experiment design on the hyperparameters, which involves computing the fisher matrix. But some other important things are:

- The plots look nicer, but this is ongoing. 
- The current main branch doesn't work with sirepo-bluesky, but this branch does. The next PR will be CI so that we don't merge things after beamline experiments that don't work with sirepo.
- Fixed to work with the updated gpytorch linear operator method as of 1.9.1 (another reason for CI)

This code is not in a great state, but it's an improvement.


mrakitin (approved):
Looks great for now!

We have a couple of outstanding issues which will be addressed via separate PRs:
- #10
- #11

PR #16, opened 2023-03-14
Add a workflow to publish the documentation

thomaswmorris:
Seems like it does


mrakitin (approved):
I added a commit to extend the existing workflow so that there is much less repetition of the same code.

Issue #10, opened 2023-03-10
Read the objects values using `.read()`


mrakitin:
This should work equally well for both `ophyd.EpicsMotor` and regular ophyd objects (`Signal`, `EpicsSignal`, etc.).

```py
In [22]: obj_em
Out[22]: EpicsMotorWithLimits(prefix='IOC:m2', name='sample_stage_x', parent='sample_stage', settle_time=0.0, timeout=None, read_attrs=['user_readback', 'user_setpoint'], configuration_attrs=['user_offset', 'user_offset_dir', 'velocity', 'acceleration', 'motor_egu', 'hlm', 'llm'])

In [23]: obj_em.read()[obj_em.name]["value"]
Out[23]: 0.0

In [24]: obj
Out[24]: EpicsSignal(read_pv='IOC:m2.VAL', name='obj', value=0.0, timestamp=1678416299.176476, tolerance=0.001, auto_monitor=False, string=False, write_pv='IOC:m2.VAL', limits=False, put_complete=False)

In [25]: obj.read()[obj.name]["value"]
Out[25]: 0.0
```

PR #18, opened 2023-03-20
dof.read() implementation

thomaswmorris:
This solves issue #10.


mrakitin (commented):
@thomaswmorris, all looks good, but could you please remove the last empty cell so that the notebook for the 2D case matches the one with the 4D one? Thanks!

thomaswmorris:
Should be good to go now!

PR #19, opened 2023-03-22
Implement non-greedy learning strategies

thomaswmorris:
These make situations with costly data acquisition (e.g. moving motors at the beamline) MUCH more efficient, and also allow us to plan our data acquisition farther in advance.


mrakitin:
#18 is merged now, I think it's worth rebasing of this PR on the `main` branch.

bloptools/gp/__init__.py
```diff
@@ -12,208 +11,99 @@
...
-
-        self.state_dict = self.model.state_dict()
-
-    def regress(self, x):
-        x = torch.as_tensor(np.atleast_2d(x)).float()
+def load(filepath, **kwargs):
+    with h5py.File(filepath, "r") as f:
+        params = f["params"][:]
+    data = pd.read_hdf(filepath, key="data")
+    return Optimizer(init_params=params, init_data=data, **kwargs)
```
mrakitin, line 23: Where do we get `Optimizer` from? Should it be imported? Or is it defined in this module?

mrakitin (approved):
I skimmed through the changes, and they look good to me. I had a single question regarding the `Optimizer` class, but it can be addressed via the next PR if there any changes are needed.

bloptools/gp/__init__.py
mrakitin: @thomaswmorris, could you please add a unit test for this load function?

PR #20, opened 2023-03-29
Add blockwise inversion and other fun stuff

thomaswmorris:
There are quite a few things in here, but they are all making it more efficient when the agent is deciding where to sample.

### DIFF
--- a/examples/prepare_gp_optimizer.py
+++ b/examples/prepare_gp_optimizer.py
@@ -42,7 +42,7 @@ root_dir = "/tmp/sirepo-bluesky-data"
 _ = make_dir_tree(datetime.datetime.now().year, base_path=root_dir)
 
 
-connection = SirepoBluesky("http://localhost:8001")
+connection = SirepoBluesky("http://localhost:8000")
 
 data, schema = connection.auth("shadow", "00000002")
 classes, objects = create_classes(connection.data, connection=connection)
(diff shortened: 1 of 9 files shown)


examples/prepare_gp_optimizer.py
```diff
@@ -42,7 +42,7 @@
 _ = make_dir_tree(datetime.datetime.now().year, base_path=root_dir)
 
 
-connection = SirepoBluesky("http://localhost:8000")
+connection = SirepoBluesky("http://localhost:8001")
```
mrakitin, line 45: Maybe instead we'd like to configure tiled to use a non-standard port, following the instructions at https://blueskyproject.io/tiled/how-to/profiles.html#create-and-use-a-profile. I think we can use `localhost:8001` for tiled rather than for Sirepo as it is already used in CI, etc.

mrakitin:
I wonder if the CI did not run. Let's "power-cycle" the PR to trigger the CI runs.

examples/prepare_gp_optimizer.py
thomaswmorris: Yeah we should do this at some point

mrakitin, line 45:
```suggestion
connection = SirepoBluesky("http://localhost:8000")
```

PR #21, opened 2023-04-11
Integrated BoTorch

thomaswmorris:
A lot of changes! This diverged a lot from #20, so I made a separate PR. The main changes are: 

- Added an experiments submodule, for easy switching between different beamlines/simulations/optimization problems.
- The optimizer is now built on Botorch models, which work with fancy information-theory acquisition functions like entropy search.
- Implemented multitask optimization, so the model can know about several experiment outputs instead of needing to collapse it to one fitness function.


mrakitin:
@thomaswmorris, I've submitted a PR to your PR branch: https://github.com/thomaswmorris/bloptools/pull/1. That should resolve the docs building issues.

mrakitin (approved):
I skimmed through this PR quickly, looks good - tests passed and the docs are building now.

PR #23, opened 2023-04-17
Restructure the BoTorch interface

thomaswmorris:
I restructured and renamed a bunch of stuff, with the philosophy that:

- The normalization of the inputs and targets happens in models.py, so that all references to "params" and "X" have units (e.g. mm for a motor) and are never normalized, which makes things a lot less confusing
- The Bayesian agent shouldn't have any torch in it, and just asks a model from models.py about the Gaussian Process.

### DIFF
--- a/bloptools/experiments/tests/__init__.py
+++ b/bloptools/experiments/tests/__init__.py
@@ -8,7 +8,7 @@ def get_dofs(n=2):
 
 
 class BaseOptimizationTest:
-    MIN_SNR = 1e2
+    MIN_SNR = 1e3
 
     n_dof = 2
     dofs = get_dofs(n=2)
@@ -64,7 +64,7 @@ class Himmelblau(BaseOptimizationTest):
 
 class Rosenbrock(BaseOptimizationTest):
     """
-    Ackley's function in $n$ dimensions (https://en.wikipedia.org/wiki/Ackley_function)
+    Rosenbrock function in $n$ dimensions (hhttps://en.wikipedia.org/wiki/Rosenbrock_function)
     """
 
     def __init__(self, n_dof=2):
@@ -78,9 +78,21 @@ class Rosenbrock(BaseOptimizationTest):
         return -np.log((100 * (X[..., 1:] - X[..., :-1] ** 2) ** 2 + (1 - X[..., :-1]) ** 2).sum(axis=-1))
 
 
+class ConstrainedRosenbrock(Rosenbrock):
+    """
+    Rosenbrock function in $n$ dimensions (hhttps://en.wikipedia.org/wiki/Rosenbrock_function)
+    """
+
+    @staticmethod
+    def fitness_func(X):
+        if (X**2).sum(axis=-1) > len(X):
+            return np.nan
+        return -np.log((100 * (X[..., 1:] - X[..., :-1] ** 2) ** 2 + (1 - X[..., :-1]) ** 2).sum(axis=-1))
+
+
 class StyblinskiTang(BaseOptimizationTest):
     """
-    Ackley's function in $n$ dimensions (https://en.wikipedia.org/wiki/Ackley_function)
+    Styblinski-Tang function in $n$ dimensions (https://en.wikipedia.org/wiki/Test_functions_for_optimization)
     """
 
     def __init__(self, n_dof=2):
--- a/bloptools/experiments/tests/__init__.py
+++ b/bloptools/experiments/tests/__init__.py
@@ -7,53 +7,88 @@ def get_dofs(n=2):
     return [Signal(name=f"x{i+1}", value=0) for i in range(n)]
 
 
-class Ackley:
+class BaseOptimizationTest:
+    MIN_SNR = 1e2
+
+    n_dof = 2
+    dofs = get_dofs(n=2)
+    bounds = np.array([[-5.0, +5.0], [-5.0, +5.0]])
+
+    DEPENDENT_COMPONENTS = [f"x{i+1}" for i in range(2)]
+
+    def initialize(self):
+        yield from bps.null()  # do nothing
+
+    def parse_entry(self, entry):
+        X = np.array([getattr(entry, cpt) for cpt in self.DEPENDENT_COMPONENTS])
+        fitness = self.fitness_func(X)
+        return ("fitness"), (fitness)
+
+
+class Ackley(BaseOptimizationTest):
     """
     Ackley's function in $n$ dimensions (https://en.wikipedia.org/wiki/Ackley_function)
     """
 
-    MIN_SNR = 1e1
-
     def __init__(self, n_dof=2):
         self.n_dof = n_dof
-        self.dofs = [getattr(self.dofs_256, f"x{i+1}") for i in range(self.n_dof)]
-        self.bounds = np.array([[0.0, 1.0] for i in range(self.n_dof)])
-
+        self.dofs = get_dofs(n=n_dof)
+        self.bounds = np.array([[-5.0, +5.0] for i in range(self.n_dof)])
         self.DEPENDENT_COMPONENTS = [f"x{i+1}" for i in range(self.n_dof)]
 
-    def initialize(self):
-        yield from bps.null()  # do nothing
-
-    def parse_entry(self, entry):
-        # get the ingredient from our dependent variables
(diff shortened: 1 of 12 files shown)


bloptools/experiments/tests/__init__.py
```diff
@@ -64,30 +78,42 @@ def fitness_func(X):
...
+        self.DEPENDENT_COMPONENTS = [dof.name for dof in self.dofs]
+
+    @staticmethod
+    def fitness_func(X):
+        return -np.log((100 * (X[..., 1:] - X[..., :-1] ** 2) ** 2 + (1 - X[..., :-1]) ** 2).sum(axis=-1))
+
+
+class ConstrainedRosenbrock(Rosenbrock):
+    """
+    Rosenbrock function in $n$ dimensions (hhttps://en.wikipedia.org/wiki/Rosenbrock_function)
```
mrakitin, line 97:
```suggestion
    Rosenbrock function in $n$ dimensions (https://en.wikipedia.org/wiki/Rosenbrock_function)
```

bloptools/experiments/tests/__init__.py
```diff
@@ -64,30 +78,42 @@ def fitness_func(X):
 
 class Rosenbrock(BaseOptimizationTest):
     """
-    Ackley's function in $n$ dimensions (https://en.wikipedia.org/wiki/Ackley_function)
+    Rosenbrock function in $n$ dimensions (hhttps://en.wikipedia.org/wiki/Rosenbrock_function)
```
mrakitin, line 81:
```suggestion
    Rosenbrock function in $n$ dimensions (https://en.wikipedia.org/wiki/Rosenbrock_function)
```

bloptools/experiments/tests/__init__.py
```diff
@@ -1,20 +1,34 @@
...
+    components = {}
+
+    for i in range(n):
+        components[f"{i+1}"] = Cpt(Signal, value=i + 1)
+
+    cls = type("DOF", (Device,), components)
+
+    device = cls(name=name)
+
+    return [getattr(device, attr) for attr in device.read_attrs]
```
mrakitin, line 21: Thanks for implementing it! Why not return the generated device?
Also, is it used anywhere? I see only the `get_dofs()` is used. I think it's fine if you only need a list of individual signals.

bloptools/experiments/tests/__init__.py
```diff
@@ -1,20 +1,34 @@
...
 
 def get_dofs(n=2):
     return [Signal(name=f"x{i+1}", value=0) for i in range(n)]
 
 
+def get_device(name="dofs", n=2):
+    components = {}
+
+    for i in range(n):
+        components[f"{i+1}"] = Cpt(Signal, value=i + 1)
```
mrakitin, line 15: We need to have the component attribute names starting with `x`, otherwise one will get this error:
```py
In [3]: get_device()
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In [3], line 1
----> 1 get_device()

Cell In [2], line 7, in get_device(name, n)
      4 for i in range(n):
      5     components[f"{i+1}"] = Cpt(Signal, value=i + 1)
----> 7 cls = type("DOF", (Device,), components)
      9 device = cls(name=name)
     11 return [getattr(device, attr) for attr in device.read_attrs]

File ~/anaconda3/envs/sirepo-bluesky/lib/python3.9/site-packages/ophyd/device.py:982, in Device.__init_subclass__(cls, **kwargs)
    980 "This is called automatically in Python for all subclasses of Device"
    981 super().__init_subclass__(**kwargs)
--> 982 cls._initialize_device()

File ~/anaconda3/envs/sirepo-bluesky/lib/python3.9/site-packages/ophyd/device.py:960, in Device._initialize_device(cls)
    957 cls.component_names = tuple(cls._sig_attrs)
    959 # The namedtuple associated with the device
--> 960 cls._device_tuple = namedtuple(
    961     f"{cls.__name__}Tuple",
    962     [comp for comp in cls.component_names[:254] if not comp.startswith("_")],
    963 )
    965 # List the attributes that are Devices (not Signals).
    966 # This list is used by stage/unstage. Only Devices need to be staged.
    967 cls._sub_devices = [
    968     attr for attr, cpt in cls._sig_attrs.items() if cpt.is_device
    969 ]

File ~/anaconda3/envs/sirepo-bluesky/lib/python3.9/collections/__init__.py:390, in namedtuple(typename, field_names, rename, defaults, module)
    388     raise TypeError('Type names and field names must be strings')
    389 if not name.isidentifier():
--> 390     raise ValueError('Type names and field names must be valid '
    391                      f'identifiers: {name!r}')
    392 if _iskeyword(name):
    393     raise ValueError('Type names and field names cannot be a '
    394                      f'keyword: {name!r}')

ValueError: Type names and field names must be valid identifiers: '1'
```
```suggestion
        components[f"x{i+1}"] = Cpt(Signal, value=i + 1)
```

bloptools/tests/test_bayesian_shadow.py
```diff
@@ -1,10 +1,12 @@
 import numpy as np
+import pytest
 from sirepo_bluesky.sirepo_ophyd import create_classes
 
 from bloptools.bo import BayesianOptimizationAgent
 from bloptools.experiments.shadow import tes
 
 
+@pytest.mark.shadow
```
mrakitin, line 9: Are those markers used anywhere?

mrakitin (commented):
@thomaswmorris, I think it's close to completion, however, needs some work. I added a few suggestions below.

Also, can the notebook with the Himmelblau test function be improved? Right now it does not uncover all 4 optimum values:

image

Maybe having a reliable random seed and an increased number of iterations can help?

bloptools/tests/test_bayesian_shadow.py
thomaswmorris: I sometimes use them to exclude sirepo tests, I can split up the tests into sirepo and non-sirepo

bloptools/experiments/tests/__init__.py
thomaswmorris: The optimizer uses an iterable of signals (so that we can optimize many devices), so having them as one device just makes it a bit more complicated for no reason

thomaswmorris:
> @thomaswmorris, I think it's close to completion, however, needs some work. I added a few suggestions below.
> 
> Also, can the notebook with the Himmelblau test function be improved? Right now it does not uncover all 4 optimum values:
> 
> image
> 
> Maybe having a reliable random seed and an increased number of iterations can help?

I added more iterations

examples/prepare_chx_shadow.py
```diff
@@ -0,0 +1,22 @@
...
+db.reg.register_handler("SIREPO_FLYER", SRWFileHandler, overwrite=True)
+
+plt.ion()
+
+root_dir = "/tmp/sirepo-bluesky-data"
+_ = make_dir_tree(datetime.datetime.now().year, base_path=root_dir)
+
+connection = SirepoBluesky("http://localhost:8000")
+
+data, schema = connection.auth("shadow", "I1Flcbdw")
```
mrakitin, line 12: I don't think this simulation exists in the database for Sirepo started with the action from the sirepo-bluesky repo. We can add it if needed. Was it converted from the `HXV1JQ5c` SRW simulation (from [this list](https://nsls-ii.github.io/sirepo-bluesky/simulations.html#other-srw-simulations)) to the Shadow simulation?

Why not to use the existing TES Shadow simulation?

mrakitin (approved):
I like the result of optimization much better now!
image

bloptools/experiments/tests/__init__.py
mrakitin: Let's keep it for now. It's good to have a quick way to use `Device`-based components if needed in the future.

bloptools/tests/test_bayesian_shadow.py
mrakitin: I remember the markers need to be declared in the `pytest.ini` file at the top of the repo.

Example 1: https://github.com/NSLS-II/sirepo-bluesky/blob/main/pytest.ini.
Example 2: https://github.com/Synchrotron-Radiation-Workshop/SRW/blob/reorg/clients/python/pytest.ini

mrakitin:
I've opened 2 issues to cover non-blocking requests raised during this PR review:
- https://github.com/NSLS-II/bloptools/issues/24
- https://github.com/NSLS-II/bloptools/issues/25

@thomaswmorris, please address them when you have time. I am going to merge this PR.

PR #26, opened 2023-04-25
Release notes and sundries for bloptools

thomaswmorris:
This PR is to ready us for a release, and it also has a few subtle tweaks.


docs/source/release-history.rst
```diff
@@ -3,6 +3,14 @@ Release History
 ===============
 
 
+v0.2.0 (2023-03-25)
```
mrakitin, line 6:
```suggestion
v0.2.0 (2023-04-25)
```

pytest.ini
```diff
@@ -0,0 +1,5 @@
+[pytest]
+markers =
+    shadow: optimization with Shadow backend (quicker simulations)
+    srw: optimization with SRW backend (slower simulations)
+    test-func: optimization on test functions (no latency)
```
mrakitin, line 5: Should it have an underscore rather than `-`, as it's used in `bloptools/tests/test_bayesian_himmelblau.py` (`@pytest.mark.test_func`)? Or does pytest understand and convert them automatically?
```suggestion
    test_func: optimization on test functions (no latency)
```

mrakitin (commented):
A couple of suggestions before it gets merged.

docs/source/release-history.rst
thomaswmorris: oops!

pytest.ini
thomaswmorris: good point, let's make them all with a `-`

thomaswmorris: oh flake really doesn't like that

PR #32, opened 2023-05-24
Simplify the syntax on initializing the agent

thomaswmorris:
A lot of the code was copied by predefining tasks in the submodules for every optimization problem. The code is now much more easily customized.


mrakitin (approved):
Looks good to me. We'll try to apply it at ISS.

PR #45, opened 2023-09-05
Added hypervolume acquisition and constraints

thomaswmorris:
Optimization constraints in botorch are weird, so we should just implement our own acquisition functions that we can manually adjust using the output of the classifier.


requirements.txt
mrakitin, line 1: Can we unpin `bluesky` as well, but add `zict<3.0.0` like in https://github.com/bluesky/bluesky/blob/6e3c4721ce728bcea2d2cc64f3e56d7c6b31e5e5/requirements.txt#L12C1-L12C11?

mrakitin (commented):
I skimmed through the PR, looks good!
We need to unpin bluesky to be able to deploy on the floor with bluesky v1.10.0.

mrakitin (approved):

PR #46, opened 2023-09-11
Tweaks to the code from work done at ATF 

thomaswmorris:
This adds e.g. an optional trigger delay to let the beam feedback system update


mrakitin:
The commits were included in #45.

PR #48, opened 2023-09-28
Monte Carlo acquisition functions

thomaswmorris:
This PR adds more Monte Carlo acquisition functions, which should be preferred to analytic ones (as they are faster to optimize and more robust).


thomaswmorris:
The `learn` method does the same thing as the `initialize` method, so it makes sense to combine them and just warn the user that they have to start with a quasi-random acquisition function.

bloptools/bayesian/__init__.py
```diff
@@ -288,13 +247,153 @@ def tell(self, new_table=None, append=True, train=True, **kwargs):
...
             except botorch.exceptions.errors.ModelFittingError:
-                if self._initialized:
+                if self.initialized:
                     self._set_hypers(cached_hypers)
                 else:
                     raise RuntimeError("Could not fit model on initialization!")
 
         self.constraint = GenericDeterministicModel(f=lambda x: self.classifier.probabilities(x)[..., -1].squeeze(-1))
 
+    def ask(self, acq_func_identifier="qei", n=1, route=True, sequential=True, **acq_func_kwargs):
```
mrakitin, line 257: I am thinking about the `acq_func_identifier` values: maybe we can define an Enum class with the values and access it via `<Class>.qei`? Maybe even the dataclass with the fields to match the ones from the config file?

That way we (1) can easily get the parameters programmatically, (2) can have it in the docs, and (3) can avoid mistyped values by validating input before passing the input to `get_acquisition_function`.

bloptools/bayesian/__init__.py
```diff
@@ -288,13 +247,153 @@ def tell(self, new_table=None, append=True, train=True, **kwargs):
...
+        """
+        Ask the agent for the best point to sample, given an acquisition function.
+
+        acq_func_identifier: which acquisition function to use
+        n: how many points to get
```
mrakitin, line 262: Let's make it `numpydoc`-compatible: https://numpydoc.readthedocs.io/en/latest/format.html#parameters.

bloptools/tests/test_acq_funcs.py
mrakitin, line 1: Can we try to cover all individual acquisition functions? I think it can be done efficiently with https://docs.pytest.org/en/7.3.x/how-to/parametrize.html.

bloptools/bayesian/__init__.py
```diff
@@ -288,13 +247,153 @@ def tell(self, new_table=None, append=True, train=True, **kwargs):
...
+        data=None,
+        **kwargs,
+    ):
+        """
+        This iterates the learning algorithm, looping over ask -> acquire -> tell.
+        It should be passed to a Bluesky RunEngine.
+        """
+
+        if data is not None:
+            if type(data) == str:
```
mrakitin, line 371:
```suggestion
            if type(data) is str:
```

mrakitin: Should we also check that the data is a valid file path?

mrakitin (commented):
Good improvement overall. I have a few suggestions below to document the code better and have better test coverage.

bloptools/bayesian/__init__.py
thomaswmorris: I think `pd.read_hdf` on the next line should handle the error ok?

bloptools/bayesian/__init__.py
thomaswmorris: This is a good idea, but I don't quite understand what you mean by using Enum here. Let's chat in person and work it out.

bloptools/bayesian/acquisition/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+
+    if acq_func_name is None:
+        raise ValueError(f'Unrecognized acquisition function identifier "{acq_func_identifier}".')
+
+    return acq_func_name
+
+
+def get_acquisition_function(agent, acq_func_identifier="qei", **acq_func_kwargs):
+    """
+    Generates an acquisition function from a supplied identifier.
```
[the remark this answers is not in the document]
thomaswmorris: I'll add docs to the next PR, it may not be worth it to do it here and deal with merge conflicts.

bloptools/bayesian/__init__.py
mrakitin: Discussed, we'll consider it after the typing annotations are implemented.

mrakitin (approved):
Looks good! The docs updates will be covered by #50.

PR #51, opened 2023-10-26
Added targeting mode to objectives

thomaswmorris:
This lets you optimize for a specific value of an objective, instead of just being able to minimize/maximize it. This is useful for e.g. getting a particular beam spot size on demand, or for "color on command" for quantum dot experiments.


thomaswmorris:
This has diverged a bit, it's easier to make a new one

PR #55, opened 2024-01-08
Add napari viewer

thomaswmorris:
A preliminary implementation of napari for blop, so that we can better see fitted models and acquisition for high-dimensional (D > 2) problems.


mrakitin:
Thanks, @thomaswmorris! A couple of snapshots would help to illustrate the interface. Could you add them? Maybe even adding them to the documentation in a dedicated section will be best.

danielballan:
Interesting. I’d be excited to see screenshots.

PR #57, opened 2024-01-12
More user-friendly controls

thomaswmorris:
Some improvements to the UX:

- Separate "search" bounds (where to look) and "trust" bounds (where to use/ignore data from). 
- Can now target an interval for the value of some objective
- Log transforms in DOFs
- Targeting constraints in objectives, so that the agent aims for some interval.


mrakitin (approved):
Looks good to me, thanks, @thomaswmorris!
Just a couple of minor suggestions below. Feel free to merge after accepting them.

PR #64, opened 2024-04-03
Add support for discrete DOFs

thomaswmorris:
They don't do anything right now, but changing a bit of syntax for later on.

PR #66, opened 2024-04-15
Improve DOFs and Objectives for better control of Pareto optimization

thomaswmorris:
This PR does two main things:

1. We interpret objectives as either a fitness (to be minimized or maximized) or a constraint (to satisfy some condition). Optimality is then the optimal/Pareto-efficient fitness within all the constriants.
2. We generalize the notion of a transform, so that we can transform the inputs and outputs with any function we want to better describe them using a GP.

Some small things:
- A lot of processes are more robust and numerically stable.
- We delete a lot of spaghetti code.
- Add more to the documentation.

### DIFF
--- a/src/blop/objectives.py
+++ b/src/blop/objectives.py
@@ -13,23 +13,23 @@ DEFAULT_MAX_NOISE_LEVEL = 1e0
 
 OBJ_FIELD_TYPES = {
     "description": "object",
-    "kind": "str",
+    # "kind": "str",
     "type": "str",
     "target": "object",
-    "active": "bool",
     "transform": "str",
+    "domain": "str",
     "trust_domain": "object",
-    "active": "bool",
-    "weight": "bool",
+    "weight": "float",
     "units": "object",
     "noise_bounds": "object",
     "noise": "float",
-    "n": "int",
+    "n_valid": "int",
     "latent_groups": "object",
+    "active": "bool",
 }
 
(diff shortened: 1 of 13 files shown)


mrakitin:
Please rebase on `main` to eliminate the commits from already merged #64.

src/blop/objectives.py
```diff
@@ -236,66 +324,86 @@ def summary(self) -> pd.DataFrame:
...
-    @property
-    def weights(self) -> np.array:
+    def transform(self, Y):
         """
-        Returns an array of the objective weights.
+        Transform the experiment space to the model space.
         """
-        return np.array([obj.weight for obj in self.objectives])
+        if Y.shape[-1] != len(self):
+            raise ValueError()
```
mrakitin, line 392: We need a message for the exception here.

src/blop/objectives.py
```diff
@@ -236,66 +324,86 @@ def summary(self) -> pd.DataFrame:
...
-        Returns a signed array of the objective weights.
+        Transform the model space to the experiment space.
         """
-        return np.array([(1 if obj.target == "max" else -1) * obj.weight for obj in self.objectives])
-
-    def add(self, objective):
-        _validate_objectives([*self.objectives, objective])
-        self.objectives.append(objective)
+        if Y.shape[-1] != len(self):
+            raise ValueError()
```
mrakitin, line 404: Please add a meaningful message here.

mrakitin (commented):
Looks good overall, but it needs a few minor touch-ups suggested below.

PR #69, opened 2024-07-07
Add simulated hardware

thomaswmorris:
Inspired by conversations with some beamline scientists, it would help to have more concrete optimization problems. This adds a lightweight simulation of a beamline mimicking a KB mirror endstation. We may add more examples in the future.


mrakitin (changes requested):
Looks good overall, but needs some cleanup (details below).

PR #73, opened 2024-07-30
Added method to prune bad data

megha-narayanan:
adds pruning method


mrakitin:
Thanks for the contribution, @megha-narayanan!

mrakitin (approved):

PR #77, opened 2024-11-18
Adding torch.tensor(np.nan) bug fix

### LINKED ISSUE #76: User report
@lyang11973 reported an issue with this optimization problem:

```py
pos_x10 = crl.x1.position
pos_x20 = crl.x2.position
pos_y10 = crl.y1.position
pos_y20 = crl.y2.position
search_range = 0.6

dofs = [
    DOF(device=crl.x1, description="transfocator upstream x", 
        search_domain=(pos_x10-search_range/2, pos_x10+search_range/2)),
    DOF(device=crl.x2, description="transfocator downstream x", 
        search_domain=(pos_x20-search_range/2, pos_x20+search_range/2)),
    DOF(device=crl.y1, description="transfocator upstream y", 
        search_domain=(pos_y10-search_range/2, pos_y10+search_range/2)),
    DOF(device=crl.y2, description="transfocator downstream y", 
        search_domain=(pos_y20-search_range/2, pos_y20+search_range/2))
]

objectives = [
    Objective(name="em1_sum_all_mean_value", description="beam intensity", target="max", trust_domain=(100000, np.inf))
]

dets = [em1]

agent = Agent(dofs=dofs, objectives=objectives, detectors=dets, db=db)

RE(fast_shutter_wrapper(agent.learn("qr", n=16))) 
```

It completes the first set of data collection, but then fails with an error:
```py
TypeError: where(): argument 'other' (position 2) must be Tensor, not float 
```

## Proposed fix

In [agent.py](https://github.com/NSLS-II/blop/blob/34f613bbd3a8d4cf956f23844306a79b6e5b9569/src/blop/agent.py#L964): `np.nan` needs to be converted to `torch.tensor(np.nan)`.

@jennmald, @yxrmz, please have a look. We need to:
- [ ] implement the fix
- [ ] add a test for the use case
- [ ] update the doc string to explain the `enforce_all_objectives_valid` parameter

jennmald:
Adds the fix for `torch.tensor(np.nan)` for #76 in `agent.py`
Still needs test case for this fix using Ophyd sim and the updated doc string to explain the `enforce_all_objectives_valid` parameter.

Closes #76


jennmald:
@mrakitin @yxrmz I made the fix for the open issue but am having trouble coming up with a test case still using Ophyd sim. Let me know if either of you have ideas. thanks!

mrakitin:
@thomaswmorris, could you please help Jennefer with the description of the parameter?

thomaswmorris:
@jennmald you would turn on `enforce_all_objectives_valid` if you want to throw away all the beam diagnostics when one of them is bad, e.g. if the flux is invalid because it's too low and so you want to automatically throw away the size estimate. But this isn't always true, e.g. at TES where you can measure the flux separately from the size and so we would still want to fit to the flux even if the size diagnostic is messed up.

jennmald:
Thanks @thomaswmorris !

PR #78, opened 2024-11-21
Refactor constraints

thomaswmorris:
This PR adds two important features:
- You can have an objective that has both a target and a constraint, e.g. "I want my beam to be as small as possible, but definitely at least this small".
- You can use only constraints "I just want my beam to be in this spot and at least this small, otherwise I don't care"

### DIFF
--- a/src/blop/agent.py
+++ b/src/blop/agent.py
@@ -17,8 +17,6 @@ import numpy as np
 import pandas as pd
 import scipy as sp
 import torch
-
-# from botorch.utils.transforms import normalize
 from botorch.acquisition.objective import ScalarizedPosteriorTransform
 from botorch.models.deterministic import GenericDeterministicModel
 from botorch.models.model_list_gp_regression import ModelListGP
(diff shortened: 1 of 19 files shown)


jennmald (approved):

src/blop/agent.py
```diff
@@ -184,7 +182,7 @@ def refresh(self):
     def redigest(self):
         self._table = self.digestion(self._table, **self.digestion_kwargs)
 
-    def sample(self, n: int = DEFAULT_MAX_SAMPLES, method: str = "quasi-random") -> torch.Tensor:
+    def sample(self, n: int = DEFAULT_MAX_SAMPLES, normalize: bool = False, method: str = "quasi-random") -> torch.Tensor:
```
mrakitin, line 185: Please add the `normalize` parameter into the docstring.

mrakitin (approved):
Looks good, went through that with @thomaswmorris interactively.
Just a couple of minor comments below.

PR #80, opened 2025-01-02
fix fitness and constraint plots

jennmald:
- [x] Fix the fitness plot issue, plots do not appear when they are supposed to
- [x] add in the constraint plots using log_total_constraints

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -83,7 +83,7 @@ jobs:
       - name: Build Docs
         run: make -C docs/ html
 
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           name: ${{ env.REPOSITORY_NAME }}-docs
           path: docs/build/html/
--- a/README.rst
+++ b/README.rst
@@ -6,11 +6,11 @@ blop
    :target: https://github.com/NSLS-II/blop/actions/workflows/testing.yml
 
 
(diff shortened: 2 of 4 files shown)


jennmald:
@thomaswmorris I fixed up the broken plots and tried to follow your advice on the constraint plots. There are some plots that look right and others that definitely don't. Can you review if those plots should even be included? Let me know what you think.

thomaswmorris:
These already look a lot better. Which constraint plots are weird? Also, could you add the bad samples (maybe as a bunch of black x's) in the leftmost column?

thomaswmorris:
"Bad" points are the `nan` values in `agent.train_targets()`. The constant values are normal, when there are no bad points the agent predicts all other points to be good with probability 1.

jennmald:
@thomaswmorris if there are nan values in` agent.train_targets()` the `_nanquantile` function makes all points in the scatter plot not visible. There is no way I have found to determine which points correlate to `np.nan` values. 
For example:
I made `values` the same as in the test but then injected a `np.nan` value at the end of the list for the last point. When the code hits:
```
values = obj._untransform(targets)
val_vmin, val_vmax = np.nanquantile(values, q=[0.01, 0.99])`
val_norm = (
            mpl.colors.LogNorm(val_vmin, val_vmax) if obj.transform == "log" else mpl.colors.Normalize(val_vmin, val_vmax)
        )
```
then the` val_norm` and everything after that is np.nan, so no values appear on the plot. Any advice on how to approach this?

PR #86, opened 2025-02-06
Refactor: Make agent default compatible with Bluesky Adaptive

maffettone:
This does a lot of re-factoring, all with the goal of not augmenting existing functionality at all. 

The Agent was split into a `BaseAgent` that it inherits from and `Agent`. Then a `BlueskyAdaptiveAgent` was built that inherits from both this `BaseAgent` and `BlueskyAdaptiveBaseAgent`, making as much use of the initial functionality as possible. 

In a nutshell, we needed to remove extra args and kwargs from the signature of data movement functions (ask/tell). The objects there, should be attributes of the Agent that can be adjusted via the REST API.

Other components of refactoring:
  - Fix ask in bluesky
  - move original tell to base class, and add bluesky tell
  - refactor out `update_models` from base tell
  -  refactor private training methods into base class, and their deps (essentially checking anything `self.` in these functions that wasn't declared)
  - Add an `unpack_run` that attempts `learn`'s approach to pulling data, but will be tiled/dbv2 compat. Digestion is risky, but I tried my best. 
  - Add measurement plan that is QServer compat (Json serializable name, args, kwargs)
  - Make bluesky relevat attrs into registered properties, and add server registrations.


maffettone:
Should any of this CI be passing? I couldn't get any tests to pass on my main branch locally. Just want to sanity check.

pyproject.toml
```diff
@@ -7,27 +7,29 @@ name = "blop"
...
-    "torch",
+  "area-detector-handlers",
+  "bluesky",
+  "bluesky-adaptive[all]",
+  "botorch",
+  "databroker",
+  "gpytorch",
+  "h5py",
+  "matplotlib",
+  "numpy<2",
```
thomashopkins32, line 26: I'm assuming you ran into an issue with `numpy>2`?

src/blop/agent.py
```diff
@@ -156,67 +134,231 @@ def __init__(
...
 
-    def __getattr__(self, attr):
-        acqf_config = acquisition.parse_acqf_identifier(attr, strict=False)
-        if acqf_config is not None:
-            acqf, _ = _construct_acqf(agent=self, acqf_name=acqf_config["name"])
-            return acqf
-        raise AttributeError(f"No attribute named '{attr}'.")
+        # check that inputs values are inside acceptable values
+        valid = (raw_inputs >= dof._trust_domain[0]) & (raw_inputs <= dof._trust_domain[1])
+        raw_inputs = torch.where(valid, raw_inputs, np.nan)
```
thomashopkins32, line 156: This is already done in `dof._transform`
```suggestion
```

src/blop/agent.py
```diff
@@ -156,67 +134,231 @@ def __init__(
...
+
+        targets_dict = {}
+        raw_targets_dict = self.raw_targets(**subset_kwargs)
+
+        for obj in self.objectives(**subset_kwargs):
+            y = raw_targets_dict[obj.name]
+
+            # check that targets values are inside acceptable values
+            valid = (y >= obj._trust_domain[0]) & (y <= obj._trust_domain[1])
+            y = torch.where(valid, y, np.nan)
```
thomashopkins32, line 183: This check happens in `obj._transform` as well.
```suggestion
```

src/blop/agent.py
```diff
@@ -293,101 +435,370 @@ def ask(self, acqf="qei", n=1, route=True, sequential=True, upsample=1, **acqf_k
...
+            "read_only_values": read_only_values,
+            # "posterior": p,
+        }
+        """
+
+        points: Dict = default_result.pop("points")
+        acqf_obj: List[ArrayLike] = default_result.pop("acqf_obj")
+        # Turn dict of list of points into list of consistently sized points
+        keys = list(points.keys())
+        points: List[Tuple[ArrayLike]] = list(zip(*[points[key] for key in keys]))
```
thomashopkins32, line 572: Simplify:
```suggestion
        # Turn dict of list of points into list of consistently sized points
        points: List[Tuple[ArrayLike]] = list(zip(*[value for _, value in points.items()]))
```

src/blop/agent.py
```diff
@@ -293,101 +435,370 @@ def ask(self, acqf="qei", n=1, route=True, sequential=True, upsample=1, **acqf_k
...
+            "acqf_kwargs": acqf_kwargs,
+            "duration_ms": duration,
+            "sequential": sequential,
+            "upsample": upsample,
+            "read_only_values": read_only_values,
+            # "posterior": p,
+        }
+        """
+
+        points: Dict = default_result.pop("points")
```
thomashopkins32, line 568: Add stricter typing here, if possible. It looks like we make some assumptions about `points` below other than the fact that its a dictionary.

src/blop/agent.py
```diff
@@ -293,101 +435,370 @@ def ask(self, acqf="qei", n=1, route=True, sequential=True, upsample=1, **acqf_k
...
+        run : BlueskyRun
+
+        Returns
+        -------
+        independent_var :
+            The independent variable of the measurement
+        dependent_var :
+            The measured data, processed for relevance
+        """
+        if self.digestion == default_digestion_function:
```
thomashopkins32, line 604: We can remove `default_digestion_function` as a default since it does nothing...
```suggestion
        if not self.digestion:
```

src/blop/agent.py
```diff
@@ -293,101 +435,370 @@ def ask(self, acqf="qei", n=1, route=True, sequential=True, upsample=1, **acqf_k
...
+            # Assume all raw data is available in primary stream as keys
+            return (
+                [run.primary.data[key].read() for key in self.dofs.names],
+                [run.primary.data[key].read() for key in self.objectives.names],
+            )
+        else:
+            # Hope and pray that the digestion function designed for DataFrame can handle the XArray
+            data: pd.DataFrame = self.digestion(run.primary.data.read(), **self.digestion_kwargs)
+            return [data.loc[:, key] for key in self.dofs.names], [data.loc[:, key] for key in self.objectives.names]
```
thomashopkins32, line 613: Assuming this prayer can be answered by unit tests? Even with some mock data, if its simpler that way.

thomashopkins32 (commented):
Most of it looks fine since its basically copy-pasted from the existing agent. I have some suggestions but nothing blocking.

I can approve once its been tested (and all other tests pass).

This package does not have great test coverage yet so I am a bit concerned about breaking changes.

pyproject.toml
maffettone: No, I messed up the merge-conflict during the rebase. Dropped the "<2".

maffettone:
What I have learned in this process is:
1. Bluesky-adaptive desperately needs review on https://github.com/bluesky/bluesky-adaptive/pull/43 and movement on dependency bloat. 
2. The CI here will be incompatible with data broker v2.

src/blop/agent.py
maffettone:
```suggestion
        if not self.digestion or self.digestion == default_digestion_function:
```
I'll save that for a future PR since this will default to true in other places doing the simple thing.

src/blop/agent.py
maffettone: Based on stepping back from the tiled/dbv2 requirements for the CI here, I'm going to hold on producing a test for this.  Really trying to show a skeleton for how it can be done without breaking existing workflows. 

We can definitely build up a smoke test, but I don't quite understand how data will be captured with the default plans using the queue server and tiled.

thomashopkins32 (approved):
LGTM

maffettone:
Hopefully this merge makes you a contributor.

PR #87, opened 2025-02-06
Add type hints

### LINKED ISSUE #85: Add mypy type hints
The current state of the package seems to have typing inconsistencies in a lot of places. We could also unearth some bugs or unsupported (but advertised) features through this process.

thomashopkins32:
This PR adds type hints in as many places as I could. I started with `mypy` but also included some fixes from `pyright`. I think the goal should be to use `pyright` in CI testing after we solidify the core parts of the package.

Through this effort, I discovered that there is a good amount of simplification that can be done to make the code more flexible/maintainable going forward. It's outside the scope of this PR but it can be done in concrete chunks!

Closes #85


thomashopkins32:
Well this PR ended up being a lot bigger than expected and there is a lot more typing work to do. Mypy officially passes without errors, however, a lot of functions are ignored because their signatures are untyped.

Merging this PR as-is should be fine (once all tests pass). Future & smaller typing improvements can be worked on separately.

I used this issue to learn a lot more about the codebase and try to understand all of the moving parts. I think there are a lot of opportunities to simplify the code, document the interactions of the various moving pieces, and have much broader test coverage.

I'll open tickets for each of these so we can begin to make progress here.

thomashopkins32:
Also, due to dependency issues, I've moved `BlueskyAdaptiveAgent` into the examples directory for now. Anyone wishing to create a `BlueskyAdaptiveAgent` can use this as a jumping off point.

thomashopkins32:
Now flake8 and black are conflicting with each other...we should probably switch to ruff :sweat_smile:

thomaswmorris (approved):

PR #92, opened 2025-02-19
Fix bugs in trust domain truncation

thomaswmorris:
`blop` is not truncating values that should be flagged as "bad" based on the `trust_domain` parameter of the `DOF` and `Objective` classes. This PR is a quick patch to fix that.

### DIFF
--- a/src/blop/agent.py
+++ b/src/blop/agent.py
@@ -157,14 +157,14 @@ class BaseAgent:
         Get the raw, untransformed inputs for a DOF (or for a subset).
         """
         if index is None:
-            return torch.cat([self.raw_inputs(dof.name) for dof in self.dofs(**subset_kwargs)], dim=-1)
-        return torch.tensor(self._table.loc[:, self.dofs[index].name].values, dtype=torch.double).unsqueeze(-1)
+            return torch.stack([self.raw_inputs(dof.name) for dof in self.dofs(**subset_kwargs)], dim=-1)
+        return torch.tensor(self._table.loc[:, self.dofs[index].name].values, dtype=torch.double)
 
     def train_inputs(self, index=None, **subset_kwargs):
         """A two-dimensional tensor of all DOF values."""
 
         if index is None:
-            return torch.cat([self.train_inputs(index=dof.name) for dof in self.dofs(**subset_kwargs)], dim=-1)
+            return torch.stack([self.train_inputs(index=dof.name) for dof in self.dofs(**subset_kwargs)], dim=-1)
 
         dof = self.dofs[index]
         raw_inputs = self.raw_inputs(index=index, **subset_kwargs)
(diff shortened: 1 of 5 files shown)


src/blop/agent.py
```diff
@@ -173,26 +173,20 @@ def train_inputs(self, index=None, **subset_kwargs):
...
-        values = {}
-
-        for obj in self.objectives(**subset_kwargs):
-            # return torch.cat([self.raw_targets(index=obj.name) for obj in self.objectives(**subset_kwargs)], dim=-1)
-            values[obj.name] = torch.tensor(self._table.loc[:, obj.name].values, dtype=torch.double)
-
-        return values
+        if index is None:
+            return {obj.name: self.raw_targets(obj.name) for obj in self.objectives(**subset_kwargs)}
+        return torch.tensor(self._table.loc[:, self.objectives[index].name].values, dtype=torch.double)
```
thomashopkins32, line 180: Let's avoid having two different kinds of return types since it can complicate type resolution. Can we do a separate function for raw_targets that takes an index and returns a `torch.Tensor`? No need to have this function handle both responsibilities.

src/blop/agent.py
```diff
@@ -173,26 +173,20 @@ def train_inputs(self, index=None, **subset_kwargs):
...
 
-    def train_targets(self, concatenate=False, **subset_kwargs):
+    def train_targets(self, index=None, concatenate=False, **subset_kwargs):
```
thomashopkins32, line 182: Same thing here, let's create a separate method that takes an index so as to not add complexity to the return type.

src/blop/agent.py
```diff
@@ -404,18 +414,18 @@ def ask(self, acqf="qei", n=1, route=True, sequential=True, upsample=1, **acqf_k
...
+                # if the model for any active objective mismatches the active dofs, reconstruct and train it
+                if getattr(obj, "model_dofs", {}) != set(active_dofs.names):
+                    self._try_construct_model(obj, error_on_fail=True)
                     train_model(obj.model)
 
+            # # check that all the objectives have models
+            # if not all(getattr(obj, "_model", None) is not None for obj in active_objs):
+            #     raise RuntimeError(
+            #         f"Can't construct non-trivial acquisition function '{acqf}' as the agent is not initialized."
+            #     )
```
thomashopkins32, line 427: Remove if unused
```suggestion
```

src/blop/tests/test_agents.py
```diff
@@ -33,6 +34,18 @@ def test_agent(agent, RE, db):
...
+    # dof = agent.dofs(active=True)[0]
+    # raw_x = agent.raw_inputs(dof.name).numpy()
+    # dof.trust_domain = tuple(np.nanquantile(raw_x, q=[0.1, 0.9]))
+
+    # # test trust domains for DOFs
+    # obj = agent.objectives(active=True)[0]
+    # raw_y = agent.raw_targets(index=obj.name).numpy()
+    # obj.trust_domain = tuple(np.nanquantile(raw_y, q=[0.1, 0.9]))
+
+    # RE(agent.learn("qei", n=2))
```
thomashopkins32, line 47: Is this meant to be commented out?

thomashopkins32 (changes requested):

Issue #96, opened 2025-02-24
Points outside the trust domain for both inputs & outputs aren't remove before model training.


thomaswmorris:
Setting e.g. `dof.trust_domain = (0, np.inf)` should set all negative values as `nan`.

thomashopkins32:
@thomaswmorris Was this fixed by #97?

thomaswmorris:
yep

PR #97, opened 2025-02-25
Refactor DOFs to fix trust domain behavior

thomaswmorris:
A fix for #96, and also a simplification of the `DOF` code.

### DIFF
--- a/src/blop/dofs.py
+++ b/src/blop/dofs.py
@@ -2,7 +2,7 @@ import logging
 import time as ttime
 import uuid
 from collections.abc import Iterable, Sequence
-from typing import Any, Literal, Optional, Union, cast, overload
+from typing import Any, Literal, cast, overload
 
 import numpy as np
 import pandas as pd
@@ -89,16 +89,16 @@ class DOF:
         name: str = None,
         description: str = "",
         type: Literal["continuous", "binary", "ordinal", "categorical"] = "continuous",
-        search_domain: Union[tuple[float, float], set[int], set[str], set[bool]] = (-np.inf, np.inf),
-        trust_domain: Optional[Union[tuple[float, float], set[int], set[str], set[bool]]] = None,
-        domain: Optional[Union[tuple[float, float], set[int], set[str], set[bool]]] = None,
+        search_domain: tuple[float, float] | set[int] | set[str] | set[bool] = (-np.inf, np.inf),
+        trust_domain: tuple[float, float] | set[int] | set[str] | set[bool] | None = None,
+        domain: tuple[float, float] | set[int] | set[str] | set[bool] | None = None,
         active: bool = True,
         read_only: bool = False,
-        transform: Optional[Literal["log", "logit", "arctanh"]] = None,
-        device: Optional[Signal] = None,
+        transform: Literal["log", "logit", "arctanh"] | None = None,
+        device: Signal | None = None,
         tags: list[str] = None,
         travel_expense: float = 1,
-        units: Optional[str] = None,
+        units: str | None = None,
     ):
         # these should be set first, as they are just variables
         self.name = name
@@ -184,7 +184,7 @@ class DOF:
         return f"{self.__class__.__name__}({filling})"
 
     @property
-    def search_domain(self) -> Union[tuple[float, float], set[int], set[str], set[bool]]:
+    def search_domain(self) -> tuple[float, float] | set[int] | set[str] | set[bool]:
         """
         A writable DOF always has a search domain, and a read-only DOF will return its current value.
         """
@@ -198,21 +198,21 @@ class DOF:
             return self._search_domain
 
     @search_domain.setter
-    def search_domain(self, value: Union[tuple[float, float], set[int], set[str], set[bool]]):
+    def search_domain(self, value: tuple[float, float] | set[int] | set[str] | set[bool]):
         """
         Make sure that the search domain is within the trust domain before setting it.
         """
         value = validate_set(value, type=self.type)
         trust_domain = self.trust_domain
         if is_subset(value, trust_domain, type=self.type, proper=False):
-            self._search_domain = cast(Union[tuple[float, float], set[int], set[str], set[bool]], value)
+            self._search_domain = cast(tuple[float, float] | set[int] | set[str] | set[bool], value)
         else:
             raise ValueError(
                 f"Cannot set search domain to {value} as it is not a subset of the trust domain {trust_domain}."
             )
 
     @property
-    def trust_domain(self) -> Union[tuple[float, float], set[int], set[str], set[bool]]:
+    def trust_domain(self) -> tuple[float, float] | set[int] | set[str] | set[bool]:
         """
         If _trust_domain is None, then we trust the entire domain (so we return the domain).
         """
@@ -235,7 +235,7 @@ class DOF:
                 # The search domain must stay a subset of the trust domain, so set it as the intersection.
                 self.search_domain = intersection(self.search_domain, value)
 
-        self._trust_domain = cast(Union[tuple[float, float], set[int], set[str], set[bool]], value)
+        self._trust_domain = cast(tuple[float, float] | set[int] | set[str] | set[bool], value)
 
     @property
     def domain(self) -> tuple[float, float] | set[int] | set[str] | set[bool]:
--- a/src/blop/dofs.py
+++ b/src/blop/dofs.py
@@ -235,7 +235,7 @@ class DOF:
                 # The search domain must stay a subset of the trust domain, so set it as the intersection.
                 self.search_domain = intersection(self.search_domain, value)
 
-        self._trust_domain = cast(tuple[float, float] | set[int] | set[str] | set[bool], value)
+        self._trust_domain = value
 
     @property
     def domain(self) -> tuple[float, float] | set[int] | set[str] | set[bool]:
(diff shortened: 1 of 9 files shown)


src/blop/dofs.py
```diff
@@ -228,15 +257,14 @@ def domain(self) -> Union[tuple[float, float], set[int], set[str], set[bool]]:
...
+            return self._domain
 
     def _transform(self, x: torch.Tensor, normalize: bool = True) -> torch.Tensor:
         if self.type != "continuous":
             raise ValueError("Cannot transform non-continuous DOFs.")
 
         # Since the DOF is continuous, we can safely assume the domains are tuples
-        domain = cast(tuple[float, float], self.domain)
-        x = torch.where((x > domain[0]) & (x < domain[1]), x, torch.nan)
+        x = torch.where(element_of(x, self.trust_domain), x, torch.nan)
```
thomashopkins32, line 267: If you could add the type casting within `element_of` that would be great. Otherwise, you are undoing work done in #87

src/blop/dofs.py
```diff
@@ -281,12 +309,12 @@ def summary(self) -> pd.Series:
...
-                    if attr in ["search_domain", "trust_domain"]:
-                        value = f"[{value[0]:.02e}, {value[1]:.02e}]"
-                    else:
-                        value = f"({value[0]:.02e}, {value[1]:.02e})"
+            # if attr in ["search_domain", "trust_domain", "domain"]:
+            #     if (self.type == "continuous") and not self.read_only and value is not None:
+            #         if attr in ["search_domain", "trust_domain"]:
+            #             value = f"[{value[0]:.02e}, {value[1]:.02e}]"
+            #         else:
+            #             value = f"({value[0]:.02e}, {value[1]:.02e})"
```
thomashopkins32, line 317: If not needed please remove
```suggestion
```

src/blop/tests/test_agents.py
```diff
@@ -33,10 +35,25 @@ def test_agent(agent, RE, db):
...
+    raw_y = agent.raw_targets(index=obj.name).numpy()
+    obj.trust_domain = tuple(np.nanquantile(raw_y, q=[0.2, 0.8]))
+
+    RE(agent.learn("qei", n=2))
+
     # save the data, reset the agent, and get the data back
-    agent.save_data("/tmp/test_save_data.h5")
-    agent.reset()
-    agent.load_data("/tmp/test_save_data.h5")
+    if sys.version_info.minor > 9:
```
thomashopkins32, line 53: Why is this needed?

src/blop/tests/test_agents.py
```diff
@@ -49,6 +66,7 @@ def test_agent(agent, RE, db):
 
 @pytest.mark.parametrize("agent", all_agents, indirect=True)
 def test_benchmark(agent, RE, db):
-    agent.db = db
-    per_iter_learn_kwargs_list = [{"acqf": "qr", "n": 32}, {"acqf": "qei", "n": 2, "iterations": 2}]
-    RE(agent.benchmark(output_dir="/tmp/blop", iterations=1, per_iter_learn_kwargs_list=per_iter_learn_kwargs_list))
+    if sys.version_info.minor > 9:
```
thomashopkins32, line 69: Why is this needed?

src/blop/tests/test_dofs.py
```diff
@@ -4,27 +4,27 @@
...
-        search_domain={"mango", "orange", "banana", "papaya"},
-        units="fruit",
+        name="fruit",
+        search_domain={"mango", "banana", "papaya"},
+        trust_domain={"mango", "orange", "banana", "papaya", "cantaloupe"},
     )
 
     dofs = DOFList([dof1, dof2, dof3, dof4])  # noqa
+
+    print(dofs)
```
thomashopkins32, line 30: Remove unnecessary print
```suggestion
```

src/blop/utils/sets.py
```diff
@@ -0,0 +1,79 @@
...
+        if len(s) == 2:
+            try:
+                x1, x2 = float(s[0]), float(s[1])
+                if x1 <= x2:
+                    return (x1, x2)
+            except Exception:
+                pass
+        raise ValueError(
+            f"Invalid continuous set {s}; valid continuous sets it must be a tuple of two numbers x1, x2 such that x2 >= x1"
+        )
```
thomashopkins32, line 18: Can you add the type casting here again? Also, we probably should raise an exception is the comparison doesn't work.
```suggestion
        if len(s) == 2:
            domain = cast(tuple[float, float], s)
            x1, x2 = domain
            if x1 <= x2:
                return (x1, x2)
        raise ValueError(
            f"Invalid continuous set {s}; valid continuous sets it must be a tuple of two numbers x1, x2 such that x2 >= x1"
        )
```

src/blop/utils/sets.py
```diff
@@ -0,0 +1,79 @@
+import numpy as np
+
+
+def validate_set(s, type="continuous") -> tuple:
```
thomashopkins32, line 4: Wrong return type here. It should be `Union[set, tuple[float, float]]`.

thomashopkins32 (changes requested):
I like the refactor, nice work!

I have some concerns about undoing work done in #87 though if you could please add those changes back. It took a lot of time to get those right :smile:

src/blop/tests/test_agents.py
thomashopkins32: I see this in #100 as well actually and it had to do with numpy 2.0 being used, if that gets merged first then you can remove this check and the tests should pass.

src/blop/tests/test_agents.py
thomashopkins32: I see this in #100 as well actually and it had to do with numpy 2.0 being used, if that gets merged first then you can remove this check and the tests should pass.

src/blop/dofs.py
```diff
@@ -143,39 +145,26 @@ def __post_init__(self) -> None:
...
+            self.search_domain = search_domain
 
+            if not self.read_only:
                 if not self.device:
-                    search_tensor = torch.tensor(continuous_search_domain, dtype=torch.float)
+                    search_tensor = torch.tensor(search_domain, dtype=torch.float)
                     transformed = self._transform(search_tensor)
                     center = float(self._untransform(torch.mean(transformed)))
                     self.device = Signal(name=self.name, value=center)
+
```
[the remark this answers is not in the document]
thomaswmorris: I added casting back in in the setters for the trust and search domains

src/blop/utils/sets.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    validate_set(s1, type=type)
+    validate_set(s2, type=type)
+    if type == "continuous":
+        new_min, new_max = min(s1[0], s2[0]), max(s1[1], s2[1])
+        if new_min <= new_max:
+            return (new_min, new_max)
+        return None
+    else:
+        return s1 | s2
```
thomashopkins32, line 66:
```suggestion
    if type == "continuous":
        s1 = cast(tuple[float, float], s1)
        s2 = cast(tuple[float, float], s2)
        new_min, new_max = min(s1[0], s2[0]), max(s1[1], s2[1])
        if new_min <= new_max:
            return (new_min, new_max)
        return None
    else:
        s1 = cast(set, s1)
        s2 = cast(set, s2)
        return s1 | s2
```

src/blop/utils/sets.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    validate_set(s1, type=type)
+    validate_set(s2, type=type)
+    if type == "continuous":
+        new_min, new_max = max(s1[0], s2[0]), min(s1[1], s2[1])
+        if new_min <= new_max:
+            return (new_min, new_max)
+        return None
+    else:
+        return s1 & s2
```
thomashopkins32, line 81:
```suggestion
    if type == "continuous":
        s1 = cast(tuple[float, float], s1)
        s2 = cast(tuple[float, float], s2)
        new_min, new_max = max(s1[0], s2[0]), min(s1[1], s2[1])
        if new_min <= new_max:
            return (new_min, new_max)
        return None
    else:
        s1 = cast(set, s1)
        s2 = cast(set, s2)
        return s1 & s2
```

thomashopkins32 (changes requested):
Almost there, the type casts should happen in `sets.py` rather than in `dofs.py`. These casts are meant to constrain the types that a variable _could be_ at runtime.

For example, the domain can be either a `set` or a `tuple[float, float]`. If the `type == "continuous"`, then we know for sure that the domain is a `tuple[float, float]` and so we want to cast it to a `tuple[float, float]`.

Since this is fairly repetitive code in `sets.py` and we abandoned support for Python 3.9 in #101 , we can actually use `typing.TypeGuard`s. So if you want to go this route please disregard my suggestions.

For example, we can define a separate helper function as so:
```python
def is_continuous(s: set | tuple[float, float], type: str = "continuous") -> TypeGaurd[tuple[float, float]]:
    return type == "continuous"
```


Then when you do
```python
if is_continuous(s, type):
    # s = cast(tuple[float, float], s) <----- No longer needed due to TypeGuard
    ...
```
The python builtin function `isinstance` is a `TypeGuard` as well. See https://peps.python.org/pep-0647/ for more details.

thomashopkins32 (approved):
LGTM!

PR #95, opened 2025-02-19
Ruff linter support; Removal of black, flake8, and isort

thomashopkins32:
Closes #91 and makes the code ruff compliant.

PR #100, opened 2025-03-03
Moved prepare_re_env.py back into package

### LINKED ISSUE #98: Recover`blop.utils.prepare_re_env`
Somehow `prepare_re_env.py` ended up in https://github.com/NSLS-II/blop/blob/main/docs/source/tutorials/. We should keep it as an importable module as it's used by other scripts outside of this repo. I believe it happened accidentally during the typing and ruff formatting efforts.

thomashopkins32:
Closes #98 

Originally moved in #87 to fix typing issues in notebooks but we will ignore them going forward in this case.


jennmald (approved):

thomashopkins32:
> Is 3.9 failing because of type changes?

The only difference between the installed packages in #95 and this PR was numpy switching to 2.0.2 from 1.26.4 so I think we should switch back for now.

mrakitin:
I think @thomaswmorris made it numpy version agnostic recently. Can we please work out why we have a failure with numpy 2? If the support of Python 3.9 is in question, let's remove it and have v3.10 a minimum supported version of Python. We don't use 3.9 on the floor and EOL is in October this year.

PR #101, opened 2025-03-03
Remove python 3.9 support

thomashopkins32:
Based on discussion in #100 

- Python 3.9 EOL is in October 2025
- We want to be able to support any numpy version

### DIFF
--- a/src/blop/de/de_opt_utils.py
+++ b/src/blop/de/de_opt_utils.py
@@ -193,7 +193,7 @@ def generate_hardware_flyers(
             max_velocity=max_velocity,
             min_velocity=min_velocity,
         )
-        for motor_name, vel, dist in zip(motors, velocities, dists):
+        for motor_name, vel, dist in zip(motors, velocities, dists, strict=False):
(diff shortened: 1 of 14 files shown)


src/blop/de/de_opt_utils.py
```diff
@@ -211,7 +211,7 @@ def generate_hardware_flyers(
             times_dict[motor_name] = time_
         times_list.append(times_dict)
 
-    for param, vel, time_ in zip(population, velocities_list, times_list):
+    for param, vel, time_ in zip(population, velocities_list, times_list, strict=False):
```
mrakitin, line 214: What does `strict=False` do in this case?

mrakitin (approved):
Wow, it turned out it's not one value update!
Nice job, looks good to me.

src/blop/de/de_opt_utils.py
thomashopkins32: Apparently it stops zipping after the shortest list is exhausted according to: https://peps.python.org/pep-0618/

The Ruff linter made this change automatically.

PR #108, opened 2025-03-12
Fix CI failures due to domain transforms

### LINKED ISSUE #106: Fix test_agents.py::test_agent[1d_1f]
It seems like in the most recent merge that a test is failing. Please see: https://github.com/NSLS-II/blop/actions/runs/13701341722/job/38315668899

thomaswmorris:
This fixes #106, which was a subtle bug about data type mismatches on trust domain boundaries which introduced some `nan`s. While I was bug hunting I also cleaned up some other stuff, including the plotting code.

### DIFF
--- a/src/blop/agent.py
+++ b/src/blop/agent.py
@@ -19,7 +19,6 @@ import torch
 from bluesky.run_engine import Msg
 from botorch.acquisition.acquisition import AcquisitionFunction  # type: ignore[import-untyped]
 from botorch.acquisition.objective import ScalarizedPosteriorTransform  # type: ignore[import-untyped]
-from botorch.models.deterministic import GenericDeterministicModel  # type: ignore[import-untyped]
 from botorch.models.model import Model  # type: ignore[import-untyped]
 from botorch.models.model_list_gp_regression import ModelListGP  # type: ignore[import-untyped]
 from botorch.models.transforms.input import Normalize  # type: ignore[import-untyped]
@@ -154,7 +153,11 @@ class BaseAgent:
         """
         if index is None:
             return torch.stack([self.raw_inputs(dof.name) for dof in self.dofs(**subset_kwargs)], dim=-1)
-        return torch.tensor(self._table.loc[:, self.dofs[index].name].values, dtype=torch.double)
+
+        key = self.dofs[index].name
+        if key in self._table.columns:
+            return torch.tensor(self._table.loc[:, self.dofs[index].name].values, dtype=torch.double)
+        return torch.ones(0)
 
     def train_inputs(self, index: str | int | None = None, **subset_kwargs) -> torch.Tensor:
         """
@@ -175,7 +178,9 @@ class BaseAgent:
         if index is None:
             return {obj.name: self.raw_targets_dict(obj.name)[obj.name] for obj in self.objectives(**subset_kwargs)}
         key = self.objectives[index].name
-        return {key: torch.tensor(self._table.loc[:, key].values, dtype=torch.double)}
+        if key in self._table.columns:
+            return {key: torch.tensor(self._table.loc[:, key].values, dtype=torch.double)}
+        return {key: torch.tensor([], dtype=torch.double)}
 
     def raw_targets(self, index: str | int | None = None, **subset_kwargs) -> torch.Tensor:
         """
@@ -318,24 +323,29 @@ class BaseAgent:
 
         active_dofs = self.dofs(active=True)
 
-        if method == "quasi-random":
-            X = utils.normalized_sobol_sampler(n, d=len(active_dofs))
-
-        elif method == "random":
-            X = torch.rand(size=(n, 1, len(active_dofs)))
-
-        elif method == "grid":
+        if method == "grid":
             read_only_tensor = cast(torch.Tensor, active_dofs.read_only)
             n_side_if_settable = int(np.power(n, 1 / torch.sum(~read_only_tensor)))
-            sides = [
-                torch.linspace(0, 1, n_side_if_settable) if not dof.read_only else torch.zeros(1) for dof in active_dofs
-            ]
-            X = torch.cat([x.unsqueeze(-1) for x in torch.meshgrid(sides, indexing="ij")], dim=-1).unsqueeze(-2).double()
+            grid_sides = []
+            for dof in active_dofs:
+                if dof.read_only:
+                    grid_sides.append(dof._transform(torch.tensor([dof.readback], dtype=torch.double)))
+                else:
+                    grid_side_bins = torch.linspace(0, 1, n_side_if_settable + 1, dtype=torch.double)
+                    grid_sides.append((grid_side_bins[:-1] + grid_side_bins[1:]) / 2)
+
+            tX = torch.cat([x.unsqueeze(-1) for x in torch.meshgrid(grid_sides, indexing="ij")], dim=-1).unsqueeze(-2)
+
+        elif method == "quasi-random":
+            tX = utils.normalized_sobol_sampler(n, d=len(active_dofs))
+
+        elif method == "random":
+            tX = torch.rand(size=(n, 1, len(active_dofs)))
 
         else:
             raise ValueError("'method' argument must be one of ['quasi-random', 'random', 'grid'].")
 
-        return X.double() if normalize else self.dofs(active=True).untransform(X).double()
+        return tX.double() if normalize else self.dofs.untransform(tX.double())
 
     # @property
     def pruned_mask(self) -> torch.Tensor:
@@ -387,7 +397,7 @@ class BaseAgent:
 
         if trusted.all():
             obj.validity_conjugate_model = None
-            obj.validity_probability = GenericDeterministicModel(f=lambda x: torch.ones(size=x.size())[..., -1])
+            # obj.validity_probability = GenericDeterministicModel(f=lambda x: torch.ones(size=x.size())[..., -1])
 
         else:
             dirichlet_likelihood = gpytorch.likelihoods.DirichletClassificationLikelihood(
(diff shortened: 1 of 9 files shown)


src/blop/agent.py
```diff
@@ -402,37 +411,44 @@ def _construct_model(self, obj, skew_dims: list[tuple[int, ...]] | None = None)
...
+            # if the current model matches the active dofs
+            if getattr(obj, "model_dofs", {}) == set(active_dofs.names):
+                # then we can use the current hyperparameters and just update the data
+                cached_hypers = obj.model.state_dict() if obj.model else None
+
+                logger.debug(f'{getattr(obj, "model_dofs", {}) = }')
+                logger.debug(f"{set(active_dofs.names) = }")
+                # if there aren't enough extra points to train yet
+                if n_trainable_points // self.train_every == len(obj.model.train_targets) // self.train_every:
+                    if not force_train:
```
thomashopkins32, line 443: This is my understanding of the logic but it took me some time to get there. Adding a bit more explanation could be helpful to newer contributors.

```suggestion
                # compare the new number of trainable points with the old number of train targets
                # if there aren't enough points yet and we do not force training, we skip building a model for this objective
                if not force_train and n_trainable_points // self.train_every == len(obj.model.train_targets) // self.train_every:
```

src/blop/agent.py
```diff
@@ -402,37 +411,44 @@ def _construct_model(self, obj, skew_dims: list[tuple[int, ...]] | None = None)
...
-
     def update_models(
         self,
-        train: bool | None = None,
+        force_train: bool = False,
     ) -> None:
+        """
+        We don't want to retrain the models on every call of everything, but if they are out of sync with
+        the DOFs then we should.
+        """
```
thomashopkins32, line 421: Make it clear the behavior of this parameter.
```suggestion
        """
        We don't want to retrain the models on every call of everything, but if they are out of sync with
        the DOFs then we should.
        
        Parameters
        ---------------
        force_train : bool
            Force re-training the model. If false, the model will skip re-training unless there are `self.train_every` more data points to train with.
        """
```

src/blop/agent.py
```diff
@@ -1139,13 +1148,18 @@ def plot_objectives(self, axes: tuple[int, int] = (0, 1), **kwargs) -> None:
         axes :
             A tuple specifying which DOFs to plot as a function of. Can be either an int or the name of DOFs.
         """
-        if len(self.dofs(active=True, read_only=False)) == 1:
-            if len(self.objectives(active=True, fitness=True)) > 0:
-                plotting._plot_fitness_objs_one_dof(self, **kwargs)
-            if len(self.objectives(active=True, constraint=True)) > 0:
-                plotting._plot_constraint_objs_one_dof(self, **kwargs)
+        self.update_models()
```
thomashopkins32, line 1151: I'm not sure we should be potentially re-training models here. I think it should be up to the user to decide if they want their models updated prior to plotting. They should really do this instead:
```python
agent.update_models()
agent.plot_objectives()
```

```suggestion
```

src/blop/agent.py
```diff
@@ -1157,6 +1171,8 @@ def plot_acquisition(self, acqf: str = "ei", axes: tuple[int, int] = (0, 1), **k
         axes :
             A tuple specifying which DOFs to plot as a function of. Can be either an int or the name of DOFs.
         """
+        self.update_models()
```
thomashopkins32, line 1174: Same thing here as other comment.

```suggestion
```

src/blop/objectives.py
```diff
@@ -143,16 +143,56 @@ def __init__(
...
+
+    @search_domain.setter
+    def search_domain(self, value: tuple[float, float] | set[int] | set[str] | set[bool]):
+        """
+        Make sure that the search domain is within the trust domain before setting it.
+        """
+        value = validate_set(value, type=self.type)
+        trust_domain = self.trust_domain
+        if is_subset(value, trust_domain, type=self.type, proper=False):
+            self._search_domain = cast(tuple[float, float] | set[int] | set[str] | set[bool], value)
```
thomashopkins32, line 157: Cast does nothing here. The type checker already knows the type from the method signature.
```suggestion
            self._search_domain = value
```

src/blop/objectives.py
```diff
@@ -162,25 +202,23 @@ def constrain(self, y: torch.Tensor) -> torch.Tensor:
         else:
             return torch.tensor([value in self.constraint for value in np.atleast_1d(y)])
 
+    @property
+    def all_valid(self) -> bool:
+        return not (self.validity_conjugate_model and self.validity_probability)
```
thomashopkins32, line 207: Double checking intention on logic here, maybe worth adding it as a comment with explanation.

If either the `validity_conjugate_model` or the `validity_probability` are not defined, then all are valid?

src/blop/plotting.py
```diff
@@ -13,133 +14,101 @@
...
+        3,
+        figsize=(12, 4 * len(agent.objectives)),
         sharex=True,
         constrained_layout=True,
     )
 
     agent.obj_axes = np.atleast_2d(agent.obj_axes)
 
+    print(agent.obj_axes.shape)
+
```
thomashopkins32, line 31: Remove print
```suggestion
```

src/blop/plotting.py
```diff
@@ -151,148 +120,161 @@ def _plot_objs_many_dofs(agent, axes=(0, 1), shading="nearest", cmap=DEFAULT_COL
...
     for obj_index, obj in enumerate(agent.objectives):
         targets = agent.train_targets(index=obj.name)[:, 0]
 
-        values = obj._untransform(targets)
+        values = obj._untransform(targets).numpy()
         # mask does not generate properly when values is a tensor (returns values of 0 instead of booleans)
-        values = np.array(values)
+        # values = np.array(values)
+
+        # print(values)
```
thomashopkins32, line 136: Remove
```suggestion
```

src/blop/plotting.py
```diff
@@ -151,148 +120,161 @@ def _plot_objs_many_dofs(agent, axes=(0, 1), shading="nearest", cmap=DEFAULT_COL
...
 
         val_vmin, val_vmax = np.nanquantile(values, q=[0.01, 0.99])
         val_norm = (
             mpl.colors.LogNorm(val_vmin, val_vmax) if obj.transform == "log" else mpl.colors.Normalize(val_vmin, val_vmax)
         )
 
-        obj_vmin, obj_vmax = np.nanquantile(targets, q=[0.01, 0.99])
-        obj_norm = mpl.colors.Normalize(obj_vmin, obj_vmax)
+        # obj_vmin, obj_vmax = np.nanquantile(targets, q=[0.01, 0.99])
+        # obj_norm = mpl.colors.Normalize(obj_vmin, obj_vmax)
```
thomashopkins32, line 144: Remove if not needed
```suggestion
```

src/blop/plotting.py
```diff
@@ -302,24 +284,27 @@ def _plot_objs_many_dofs(agent, axes=(0, 1), shading="nearest", cmap=DEFAULT_COL
...
-                xytext=(-ax.yaxis.labelpad - pad, 0),
-                color="k",
-                xycoords=ax.yaxis.label,
-                textcoords="offset points",
-                size="large",
-                ha="right",
-                va="center",
-                rotation=90,
-            )
+    # if len(agent.objectives) > 1:
```
thomashopkins32, line 287:
```suggestion
```

src/blop/agent.py
```diff
@@ -281,7 +286,7 @@ def fitness_scalarization(self, weights: str | torch.Tensor = "default") -> Scal
             weights *= len(active_fitness_objectives) / weights.sum()
         elif not isinstance(weights, torch.Tensor):
             raise ValueError(f"'weights' must be a Tensor or one of ['default', 'equal', 'random'], and not {weights}.")
-        return ScalarizedPosteriorTransform(weights=weights)
+        return ScalarizedPosteriorTransform(weights=weights * active_fitness_objectives.signs)
```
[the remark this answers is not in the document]
thomashopkins32: I saw the removal of the negation on targets later on in the PR, is this the only place where a change was needed to add back the negation of targets?

src/blop/agent.py
```diff
@@ -154,7 +153,11 @@ def raw_inputs(self, index: str | int | None = None, **subset_kwargs) -> torch.T
         """
         if index is None:
             return torch.stack([self.raw_inputs(dof.name) for dof in self.dofs(**subset_kwargs)], dim=-1)
-        return torch.tensor(self._table.loc[:, self.dofs[index].name].values, dtype=torch.double)
+
+        key = self.dofs[index].name
+        if key in self._table.columns:
+            return torch.tensor(self._table.loc[:, self.dofs[index].name].values, dtype=torch.double)
+        return torch.ones(0)
```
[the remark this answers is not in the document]
thomaswmorris: It handles `None` with the `if index is None`. I also think asking for a tensor all the samples when there are 0 samples and getting a list of length 0 make sense. If it causes an error anywhere it will be a shape error, which should be easy to track down.

src/blop/agent.py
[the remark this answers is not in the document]
thomaswmorris: This was intentional, it makes more sense to negate the weights (which are the better way to control the min/max behavior in the acquisition) than to negate the entire model that the agent builds.

src/blop/agent.py
```diff
@@ -402,37 +411,44 @@ def _construct_model(self, obj, skew_dims: list[tuple[int, ...]] | None = None)
...
         objectives_to_model = self.objectives if self.model_inactive_objectives else self.objectives(active=True)
+
         for obj in objectives_to_model:
-            t0 = ttime.monotonic()
+            # do we need to update the model for this objective?
+            n_trainable_points = sum(~self.train_targets(obj.name).isnan())
+
+            # if we don't have enough points
+            if n_trainable_points < 4:
+                continue
```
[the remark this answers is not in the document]
thomaswmorris: Good idea

src/blop/objectives.py
thomaswmorris: The `Objective` gets its model from the `_construct_model()` method in the `Agent`, which gives it a `validity_conjugate_model` if it detects any invalid points (otherwise it's `None`). So if this attribute is not `None` then the points are all valid. But whether the points are all valid depend on the other objectives and the `enforce_all_objectives_valid` flag, so we have to do it in this sort of roundabout way.

thomashopkins32:
The plotting functions need to remove the `update_models()` part, its not part of their responsibility to update the models.

thomashopkins32:
I'll let this PR move forward and we can decide on whether the plotting functions should update models at a later time. Discussion in #109

thomashopkins32 (approved):

PR #112, opened 2025-04-16
Ax integrations with the Bluesky ecosystem

### LINKED ISSUE #93: Ax support
It seems like https://ax.dev/ has some similar and useful abstractions we can support in blop.

Their concept of [Parameter](https://ax.readthedocs.io/en/latest/core.html#module-ax.core.parameter) maps to our `DOF` fairly cleanly.
Their concept of [Objective](https://ax.readthedocs.io/en/latest/core.html#objective) maps to our `Objective` as well.

They also have other abstractions that could be useful for tracking experiments such as [Experiment](https://ax.readthedocs.io/en/latest/core.html#module-ax.core.experiment) and [Trial](https://ax.readthedocs.io/en/latest/core.html#module-ax.core.trial).

We should investigate how we can adapt this package's use to function well with the bluesky run engine and tiled/databroker.

Possibly creating a `BlopRunner` that inherits from [Runner](https://ax.readthedocs.io/en/latest/core.html#module-ax.core.runner)?

thomashopkins32:
This PR enables using https://ax.dev/ with the Bluesky ecosystem.

- Creates some simple functions for using Ax with the Bluesky ecosystem.
- Adds a unit test using the simulated beamline.
- Adds a demo notebook showing that optimization works on the simulated KB mirror setup (see the types of visualizations Ax provides).

Closes #93

### DIFF
--- a/docs/source/tutorials/ax-blop-demo.ipynb
+++ b/docs/source/tutorials/ax-blop-demo.ipynb
@@ -323,7 +323,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.11.12"
+   "version": "3.12.8"
   }
  },
  "nbformat": 4,
--- a/docs/source/tutorials/ax-blop-demo.ipynb
+++ b/docs/source/tutorials/ax-blop-demo.ipynb
@@ -1,0 +1,331 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Simulated KB Mirror Demo using Ax and Blop\n",
+    "\n",
+    "This notebook introduces the use of the [Ax Adaptive Experimentation Platform](https://ax.dev) with integrations for Blop.\n",
+    "\n",
+    "Blop integrates the following into Ax:\n",
+    "- Running Bluesky plans using the run engine\n",
+    "- Using devices as parameters\n",
+    "- Using detectors to produce data\n",
+    "- Retrieving the results from databroker\n",
+    "\n",
+    "These features make it simple to optimize your beamline using both the [Bluesky ecosystem](https://blueskyproject.io) and Ax."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Preparing a test environment\n",
+    "\n",
+    "Here we prepare the `RunEngine` and data service `Databroker`."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from blop.utils import prepare_re_env  # noqa\n",
+    "\n",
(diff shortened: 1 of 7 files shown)


docs/source/tutorials/ax-blop-demo.ipynb
```diff
@@ -0,0 +1,331 @@
...
+    "A single Ax `Trial` represents the training and evaluation of BoTorch models to produce a suggested next `Arm`. An `Arm` in Ax is a single parameterization to be evaluated while a `Trial` can consist of many `Arm`s. In this demo, we have a single `Arm` per `Trial`."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "for _ in range(25):\n",
```
jennmald, line 190: Question about Ax, you picked 0-25 as the range for optimization, does more trials improve the optimization process or is it just like an ensemble and averages out at the end? 
I see in the markdown you mention the first few are random and rest is BO.

docs/source/tutorials/ax-blop-demo.ipynb
```diff
@@ -0,0 +1,331 @@
...
+   "outputs": [],
+   "source": [
+    "render(plot_slice(model, \"bl_kbv_dsv\", \"beam_area\"))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Viewing each arm's objective values"
```
jennmald, line 246: I am a huge fan of the visuals Ax provides.

src/blop/utils/__init__.py
```diff
@@ -17,13 +17,28 @@ def get_beam_stats(image: np.ndarray, threshold: float = 0.5) -> dict[str, float
     fim = sp.ndimage.median_filter(fim, size=3)
     fim = sp.ndimage.gaussian_filter(fim, sigma=1)
 
-    m = fim > threshold * fim.max()
+    m = fim > (threshold * fim.max())
     area = m.sum()
+    if area == 0.0:
```
jennmald, line 22: why is this needed?

docs/source/tutorials/ax-blop-demo.ipynb
thomashopkins32: Ax actually infers what optimization setup to run based on the parameters you pass in according to the docs: https://ax.readthedocs.io/en/stable/service.html#module-ax.service.ax_client.

If you want something more custom then you can pass in your own `GenerationStrategy` (which suggests new parameters, or `Arm`s, to try).

src/blop/utils/__init__.py
thomashopkins32: When I had the simulated beamline without noise and the beam was off-screen, the total sum of the image that is above the threshold of 0.5 was 0.0. This leads to a divide by zero error a few lines down. Since the image is all the same value, there is no beam to extract and we can return 0.0 for everything.

PR #114, opened 2025-05-12
Tiled server

jessica-moylan:


src/blop/sim/beamline.py
```diff
@@ -68,53 +72,72 @@ def trigger(self) -> NullStatus:
...
             getattr(self, attr).put(stats[attr])
 
         super().trigger()
 
         return NullStatus()
+    
+    def _generate_file_path(self, date_template="%Y/%m/%d"):
+        date = datetime.datetime.now()
+        assets_dir = date.strftime(date_template)
+        data_file = f"{new_uid()}.h5"
```
jennmald, line 94: This line is unused and causes pre-commit to error out, can this be removed?

src/blop/sim/beamline.py
```diff
@@ -68,53 +72,72 @@ def trigger(self) -> NullStatus:
...
         self._assets_dir = date.strftime("%Y/%m/%d")
         data_file = f"{new_uid()}.h5"
 
-        self._resource_document, self._datum_factory, _ = compose_resource(
-            start={"uid": "needed for compose_resource() but will be discarded"},
-            spec="HDF5",
-            root=self._root_dir,
-            resource_path=str(Path(self._assets_dir) / Path(data_file)),
-            resource_kwargs={},
+        uri = f"file://localhost/{str(full_path).strip('/')}"
```
jennmald, line 105: `full_path` does not exist, I would double check what this was supposed to do.

src/blop/sim/beamline.py
```diff
@@ -68,53 +72,72 @@ def trigger(self) -> NullStatus:
...
+
+        (
+            self._stream_resource_document,
+            self._stream_datum_factory,
+        ) = compose_stream_resource(
+            mimetype="application/x-hdf5",
+            uri=uri,
+            data_key=self.image.name,
+            parameters={
+                "chunk_shape": (1, *image_shape),
```
jennmald, line 115: `image_shape` is also not defined

src/blop/sim/beamline.py
```diff
@@ -68,53 +72,72 @@ def trigger(self) -> NullStatus:
...
         self._h5file_desc = h5py.File(self._data_file, "x")
         group = self._h5file_desc.create_group("/entry")
         self._dataset = group.create_dataset(
             "image",
-            data=np.full(fill_value=np.nan, shape=(1, *self.image_shape.get())),
-            maxshape=(None, *self.image_shape.get()),
-            chunks=(1, *self.image_shape.get()),
+            data=np.full(fill_value=np.nan, shape=(1, *image_shape)),
+            maxshape=(None, *image_shape),
+            chunks=(1, *image_shape),
```
jennmald, line 135: more `image_shape` is undefined

src/blop/utils/prepare_re_env.py
```diff
@@ -1,50 +1,45 @@
...
 
-    return {
-        "RE": RE,
-        "db": db,
-        "bec": bec,
-    }
+    RE.subscribe(tiled_writer) 
 
+    _ = make_dir_tree(datetime.datetime.now().year, base_path=root_dir)    
+    return dict(RE=RE, db=tiled_client, bec=bec)
```
jennmald, line 37: this can just be rewritten as:
`{RE:RE, db:tiled_client, bec:bec}`
and avoid calling dict

jennmald (changes requested):
This should help the pre-commit test run!

jennmald:
I was able to fix the docs build. the other suggestions should fix the precommit. Make sure to review the changes I made for the docs build and let me know if you have any questions!

pyproject.toml
```diff
@@ -75,6 +75,16 @@ dev = [
...
+  "stamina",
+  "uvicorn",
+  "asgi_correlation_id",
+  "fastapi",
+  "awkward",
+  "sparse",
+  "sqlalchemy",
+  "python-jose", 
+  "jmespath",
+  "uvicorn",
```
tacaswell, line 87:
```suggestion
  "pyarrow",
  "openpyxl",
```

docs/source/tutorials/introduction.ipynb
```diff
@@ -26,9 +26,34 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from blop.utils import prepare_re_env  # noqa\n",
```
jennmald, line 29: I would delete the prepare_re_env file

examples/prepare_bluesky.py
```diff
@@ -16,11 +17,11 @@
...
-    pass
+SERVER_HOST_LOCATION = "http://localhost:8000"
+tiled_client = from_uri(SERVER_HOST_LOCATION, api_key="secret")
+
+tiled_writer = TiledWriter(tiled_client)
+
+RE.subscribe(tiled_writer)
 
-RE.subscribe(db.insert)
+print(tiled_writer)
```
jennmald, line 27: is this print statement needed?

scripts/gui.py
```diff
@@ -1,27 +1,27 @@
 import asyncio
 
-import databroker
+# import databroker
```
jennmald, line 3: delete this line

jennmald (approved):

config_tiled.yml
```diff
@@ -0,0 +1,43 @@
+database:
+  uri: "sqlite+aiosqlite:///./tiled.sqlite"
+  init_if_not_exists: true
+authentication:
+  allow_anonymous_access: True
+  tiled_admins:
+  - provider: pam
```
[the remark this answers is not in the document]
genematx: I really doubt you need so involved auth here.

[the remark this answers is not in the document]
genematx: Who's Bill? :)

src/blop/de/de_optimization.py
```diff
@@ -347,8 +347,8 @@ def optimization_plan(
     bounds : dict of dicts
         Keys are motor names and values are dicts of low and high bounds. See format below.
         {'motor_name': {'low': lower_bound, 'high': upper_bound}}
-    db : databroker.Broker
-        databroker V1 instance
+    tiled : tiledWriter.Container
```
genematx, line 350: I don't think there's such a thing as `tiledWriter.Container`. `TiledWriter` is technically a Bluesky _callback_, and `Container` is an internal Tiled concept, similar to a folder or a directory.

src/blop/digestion/__init__.py
```diff
@@ -7,7 +8,17 @@ def default_digestion_function(df: pd.DataFrame) -> pd.DataFrame:
...
+    abc = xr[image_key].values
+    xr = xr.drop_vars(image_key)
+    df1 = xr.to_dataframe()
+
+    df2 = pd.DataFrame({image_key: [abc[i] for i in range(abc.shape[0])]}, index=range(abc.shape[0]))
+    df = pd.concat([df1, df2], axis=1)
+
     # Get the beam stats for each image in the dataframe and add them as new columns
     df = pd.concat([df, df[image_key].apply(lambda img: pd.Series(get_beam_stats(img, **kwargs)))], axis=1)
+    print(len(df))
```
genematx, line 23: where and when is this supposed to be printed? Maybe better to add a short message of what this number is, or put it in logs?

src/blop/integrations/ax.py
```diff
@@ -78,7 +78,7 @@ def evaluate(parameterization: dict[str, float] | dict[str, list[float]]) -> dic
         uid = RE(plan_function(readables, *unpacked))
 
         # Fetch the data
-        results_df = db[uid][0].table(fill=True)
+        results_df = db[uid]["streams", "primary", "bl_det_image"].read()
```
genematx, line 81: simply calling `.read()` maybe a bit too ambitious if the database is huge. Should be fine though if the data can be fetched quickly and fits into memory.

src/blop/sim/beamline.py
```diff
@@ -126,12 +138,23 @@ def unstage(self) -> list[Any]:
         self._datum_factory = None
         return devices
 
-    def collect_asset_docs(self) -> Generator[tuple[str, dict[str, Any]], None, None]:
+    def describe(self):
+        res = super().describe()
+        res[self.image.name] = {
```
genematx, line 143: 👍

src/blop/sim/handlers.py
```diff
@@ -23,9 +24,6 @@ class ExternalFileReference(Signal):
     def describe(self):
         resource_document_data = super().describe()
```
genematx, line 25: I think the variable names here are a bit confusing. Does this produce Descriptors or Resources? They are two distinct document types.

src/blop/sim/xrt_beamline.py
```diff
@@ -3,22 +3,24 @@
...
+from ophyd import Device, EpicsSignal, Kind, Signal  # type: ignore[import-untyped]
+from ophyd.sim import NullStatus, new_uid  # type: ignore[import-untyped]
+from ophyd.utils import make_dir_tree  # type: ignore[import-untyped]
 
 from blop.sim.handlers import ExternalFileReference
 from blop.sim.xrt_kb_model import build_beamline, build_histRGB, run_process
 from blop.utils import get_beam_stats
 
 TEST = False
+DETECTOR_STORAGE = "/tmp/blop/sim"
```
genematx, line 23: It's generally not the best practice to hard-code paths like these deep in the code. If you can set it in a notebook that calls these functions, that would be better.
Feel free to ignore this comment; I might be misunderstanding the purpose of this file.

src/blop/tests/test_agents.py
```diff
@@ -65,7 +66,7 @@ def test_agent(agent, RE, db):
 
 
 @pytest.mark.parametrize("agent", all_agents, indirect=True)
-def test_benchmark(agent, RE, db):
-    agent.db = db
+def test_benchmark(agent, RE, tiled_client):
+    agent.tiled = tiled_client
```
genematx, line 70: I've noticed there attributes in different classes called `.tiled` that are set to either TiledWriter or, like here, a tiled client. It might be worth wile making the names of those attributes more explicit (and possibly general), e.g. `.writing_callback`, `.reading_client`, or just `.tiled_writer` and `.tiled_client`. This may not make sense though -- your call.

config_tiled.yml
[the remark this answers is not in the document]
jessica-moylan: I believe this code might be artifacts from a previous version of how the code works as I used to use a custom tiled configuration.

src/blop/integrations/ax.py
jessica-moylan: I had an issue with this for the kb_mirrors and the jupyter notebook crashing due to line 16 in /src/blop/digestion/__init__.py  which is why I split the data  up. I am not quite sure what a better way of going about this might be

src/blop/sim/xrt_beamline.py
jessica-moylan: That's a good point. I will try to see if there is a way to move in further upstream.

src/blop/integrations/ax.py
genematx: oh sorry, I should have been clearer; `.read` loads the entire table or array, but if you need only a certain slice (for arrays) or columns (for tables), you can pass them directly to the array/table client, similar to pandas, e.g. `db[uid]["streams", "primary", "bl_det_image"][:1,...]`, which should be faster.

thomashopkins32 (changes requested):
Looks great! I have a few concerns before we merge:

- Why are the Xarray changes necessary? If they are, I think we should split that into a separate PR. If you can't have Tiled integration without the Xarray stuff, then let's discuss further.
- I don't think we should be dropping support for databroker. Many beamlines and other facilities which want to use Blop will depend on databroker support.
- Some other packaging related suggestions

pyproject.toml
```diff
@@ -77,6 +77,25 @@ dev = [
...
+  "pyarrow",
+  "openpyxl",
+  "numpydoc",
+  "alembic",
+  "canonicaljson",
+  "aiosqlite",
+  "duckdb",
+  "adbc-driver-manager",
+  "hdf5plugin",
+  "h5netcdf",
```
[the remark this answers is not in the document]
jessica-moylan: I am planning on solving this in the next PR with pixi

pyproject.toml
```diff
@@ -77,6 +77,24 @@ dev = [
   "pandas-stubs",
   "types-PyYAML",
   "mypy",
+  "stamina",
```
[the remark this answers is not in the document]
jessica-moylan: I am planning on solving this in the next PR with pixi

jessica-moylan:
Closed PR in favor of #143 as there was a need to keep databroker as a viable data storage option.

PR #124, opened 2025-07-25
Implement multitask models

thomaswmorris:
This PR generalizes the `LatentGP` with the `MultiTaskLatentGP`, but doesn't change any functionality. It's a starting point for introducing the option to have one model describe multiple objectives to better model them.

### DIFF
--- a/docs/wip/constrained-himmelblau copy.ipynb
+++ b/docs/wip/constrained-himmelblau copy.ipynb
@@ -28,9 +28,9 @@
    "source": [
     "import matplotlib as mpl\n",
     "import numpy as np\n",
-    "from blop.tasks import Task\n",
     "from matplotlib import pyplot as plt\n",
     "\n",
+    "from blop.tasks import Task\n",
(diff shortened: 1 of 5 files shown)


src/blop/bayesian/models.py
```diff
@@ -51,17 +54,24 @@ def construct_single_task_model(
         ),
     )
 
-    input_transform = botorch.models.transforms.input.Normalize(d=X.shape[-1])
+    input_transform = botorch.models.transforms.input.Normalize(d=X.shape[-1] + 1)
```
thomashopkins32, line 57: From what I understand about MultiTaskGPs, you should not be normalizing the task indices since they represent categorical features. To be fair, I am not aware if BoTorch internally treats the `task_feature` column as special in this regard.

I would think that you shouldn't need the extra dim here but please let me know otherwise.

thomashopkins32 (changes requested):
Two comments to address and then I can review again

Issue #116, opened 2025-06-17
Pixi has difficulties parsing current state of pyproject.toml


jessica-moylan:
Adding pixi, version 0.48.2, to the existing pyproject.toml raises a few issues as pixi is strict with parsing toml files.

Issues:
1. Pixi is unable to determine nesting keys if it is not explictly defined if there are multiple things being defined in the same layer
      This was solved by changing line 95 to `tool.hatch.build.hooks' and 96 to `vcs.version-file= "src/blop/_version.py"
       and than again adding `tool.ruff.` to any line that uses lint. This the solves the duplicate key: "build" issue. 

2. When than trying to extend the pyproject.toml file using `pixi init` and selecting **yes** to the following when the pyproject.toml file is open, when the pixi enviornment is then attempted to be built, using `pixi install` it errors with the following, ` hint: This usually indicates a problem with the package or the build environment.`
        
Regarding the second problem, there are a few suggestions of going about using pixi.
1. Keep everything in a pyproject.toml and extend the file to use pixi as well through [tool.pixi.*]
2. Move everything to a pixi.toml and remove pyproject.toml
3. Have both a pyproject.toml and pixi.toml with the pixi.toml dedicated to pixi interactions.

I am personally more of a fan of either the first two options as it will keep everything in one file but I am not quite sure what would be the best course of action when going about this.

tacaswell:
Yes, all pixi stuff should be in one or the other.  I do not know, but strongly suspect that pixi does not support reading configuration from both (if it did it would have to have rules for how to apply precedence of conflicting configuration in both which seems like a lot of complexity when "don't" is an option!)

thomashopkins32:
I've encountered this issue before I think. It's when the keys are not expanded properly, such as:
https://github.com/NSLS-II/blop/blob/main/pyproject.toml#L107-L125

Which should be expanded to
```
[tool.ruff.lint]
select = [
    "B",       # flake8-bugbear - https://docs.astral.sh/ruff/rules/#flake8-bugbear-b
    "C4",      # flake8-comprehensions - https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4
    "E",       # pycodestyle errors - https://docs.astral.sh/ruff/rules/#error-e
    "F",       # pyflakes rules - https://docs.astral.sh/ruff/rules/#pyflakes-f
    "W",       # pycodestyle warnings - https://docs.astral.sh/ruff/rules/#warning-w
[...]

PR #135, opened 2025-08-06
Adding Pixi

### LINKED ISSUE #90: Add Pixi support
https://pixi.sh/latest/

jessica-moylan:
Resolves #90 and #116 

Adds pixi as an option to run blop in addition to pip installing as well as on the CI

- The testing suite as well as the tutorials can be ran using the `dev` environment 
- pre-commit can be ran using `pixi run -e pre-commit pre-commit run --all`
- Building documentation is ran using the `docs` environment


thomashopkins32 (approved):
There are quite a few pinned dependencies. I think it's fine for now but we may have to relax those constraints in the future. Otherwise, it can be a lot to manage manually.

Otherwise, LGTM!

thomashopkins32:
@jessica-moylan I realized after the fact that this change breaks the CI testing. I am working on a PR to fix it. There are two issues I identified:
- Caching the Pixi env, which we can't do when we have blop as pip-installable
- Not actually testing different Python versions, each test is running at least Python 3.12

Issue #71, opened 2024-07-30
Make the codebase compatible with databroker v2/tiled


mrakitin:
We need to make sure it's future-proof and can work with databroker v2/tiled.
Things to address are the handlers which are registered directly with databroker, some databroker v1 specifics, etc.

whs92:
The `acquire` method passes a pd.dataframe to the `digestion` method. At the moment this uses the method `table` of the `db` object. 

Has anybody else made a merge request changing this to the V2 syntax using `to_dataframe`?

whs92:
Discussing this with Dan, I see that the problem is probably bigger

PR #143, opened 2025-08-12
Adding tiled comptability using dictionaries

jessica-moylan:
Resolves: #71 and Closes PR #114 

Currently blop runs using databroker. However, there was a want to include the ability to run tiled as well, as well as provide an opportunity to expand to allow compatibility with other data sources. When creating a RE environment, ```method``` can be defined as tiled or databroker indicating which data source will be used.
```
env = setup_re_env(db_type="temp", root_dir="/tmp/blop/sim", method="tiled")
 ```
While either can be selected, the inner working of blop runs using dictonaries, selected to limit the dependancy on an outside data source. The dictonaries are formatted as ```{key1: [list_of_values], key2: [list_of_values, ...}``` to be simple to do operations on and convert back and forth between data structures

The data is converted to use dictonaries during the ```acquire``` function and then back to the perferred datatype of the datasource, either xarray (for tiled) or pd.Dataframe (for databroker), when calling ```agent.best```

### CODE REVIEW

**Review by thomashopkins32 — CHANGES_REQUESTED** [review]
Great work so far! I have some comments related to potential bugs that I found.

Overall, I think this approach is good. However, I would prefer to abstract away the data access layer from `agent.py` altogether.

I would recommend the following:
- Create `src/blop/data_access.py`
- Define a `DataAccess` abstract base class with a method to `get()` the data.
- Implement:
    - `TiledDataAccess` which reads data from Tiled client & transforms from XArray -> dict and dict -> XArray
    - `DatabrokerDataAccess` which reads data from Databroker client & transforms from pd.DataFrame -> dict and dict -> pd.DataFrame
- Then, an `Agent` initializes a `DataAccess` class and uses that to `get()` the data for that run

Let me know if you need help with this!

File: src/blop/agent.py [hunk 1]
```diff
@@ -1065,15 +1129,53 @@ def all_acqfs(self) -> pd.DataFrame:
         """
         return acquisition.all_acqfs()
 
+    def convert_back_from_dictionary(self, data):
```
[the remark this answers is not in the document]
jennmald: it's compatibility reasons - it goes back to it's original form to avoid any conflicts.

src/blop/data_access.py
```diff
@@ -0,0 +1,84 @@
...
+                data_vars[key] = xr.DataArray(value, dims=var_dims)
+            else:
+                data_vars[key] = xr.DataArray(value)
+        return xr.Dataset(data_vars)
+
+
+class DatabrokerDataAccess(DataAccess):
+    def get_data(self, data):
+        """Retrieve data from a database based"""
+        return data.table(fill=True)
```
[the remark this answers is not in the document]
thopkins32: No longer needed, just make sure this returns a dictionary rather than a dataframe.

src/blop/integrations/ax/agent.py
```diff
@@ -59,8 +62,8 @@ class AxAgent:
         The degrees of freedom that the agent can control, which determine the output of the model.
     objectives : list[Objective]
         The objectives which the agent will try to optimize.
-    db : Broker
-        The databroker instance to read back data from a Bluesky run.
+    db : databroker or tiled
```
thopkins32, line 65: Give the object types here:
- `Broker`
- `Container` (I think)
```suggestion
    db : `Broker | Container`
```

src/blop/integrations/ax/agent.py
```diff
@@ -219,6 +222,34 @@ def _unpack_parameters(self, parameterizations: list[TParameterization]) -> list
...
+            return dictonary
+
+        elif isinstance(db, databroker.v1.Header):
+            data = db.table(fill=True)
+            for i in data:
+                dictonary[i] = data[i].to_list()
+            return dictonary
+
+        else:
+            raise ValueError("Unknown data source.")
```
thopkins32, line 251: Remove this in favor of `self.data_access` in the same pattern you use in the other Agent. This code already exists in `data_access.py` so let's re-use in this agent class as well.

src/blop/integrations/ax/agent.py
```diff
@@ -241,7 +272,7 @@ def acquire(self, trials: dict[int, TParameterization]) -> Generator[Msg, str, d
         """
         plan_args = self._unpack_parameters(trials.values())
         uid = yield from list_scan(self.readables, *plan_args, md={"ax_trial_indices": list(trials.keys())})
-        results_df = self.db[uid].table(fill=True)
+        results_df = self.convert_to_dictonary(self.db[uid])
```
thopkins32, line 275: This should be `self.data_access.get_data()`
```suggestion
        results = self.data_access.get_data(uid)
```

src/blop/integrations/ax/helpers.py
```diff
@@ -59,6 +60,35 @@ def create_bluesky_evaluator(
...
+        # if it is a databroker instance
+        elif isinstance(db, list):
+            data = db[0].table(fill=True)
+            for i in data:
+                dictonary[i] = data[i].to_list()
+            return dictonary
+
+        else:
+            raise ValueError("Unknown data source.")
+
```
thopkins32, line 91: Please use the `data_access.py` interface here instead of copying the same code again in this file.

src/blop/integrations/ax/helpers.py
```diff
@@ -78,7 +108,7 @@ def evaluate(parameterization: dict[str, float] | dict[str, list[float]]) -> dic
         uid = RE(plan_function(readables, *unpacked))
 
         # Fetch the data
-        results_df = db[uid][0].table(fill=True)
+        results_df = convert_to_dictonary(db[uid])
```
thopkins32, line 111: Same thing here, use `data_access.get_data()`
```suggestion
        results = data_access.get_data(uid)
```

src/blop/agent.py
```diff
@@ -124,18 +128,16 @@ def raw_inputs(self, index: str | int | None = None, **subset_kwargs) -> torch.T
             return torch.stack([self.raw_inputs(dof.name) for dof in self.dofs(**subset_kwargs)], dim=-1)
 
         key = self.dofs[index].name
-        if key in self._table.columns:
-            return torch.tensor(self._table.loc[:, self.dofs[index].name].values, dtype=torch.double)
+        if key in self._table:
+            return torch.tensor(self._table[self.dofs[index].name], dtype=torch.double)
```
thopkins32, line 132:
```suggestion
            return torch.tensor(self._table[key], dtype=torch.double)
```

thopkins32 (changes requested):
This is a large and complex PR in it's current state. I think it can be greatly simplified by removing the duplicate code as well as the unused methods and variables.

There are also some critical bugs that must be fixed before we can merge. I wish we had better unit tests that can easily point out issues like this, but we aren't there yet! (it would be a good idea to add these unit tests if you wish). The way you currently have it, new data is not being added at all. I am very surprised that the unit tests are not catching this...

I may have additional comments after this review, so please loop me in ahead of time if you have any questions.

Issue #140, opened 2025-08-07
Cleanup repo for new release


thomashopkins32:
- Remove outdated tool configurations
  - isort
  - flake8
  - etc.
- Remove duplicate configurations
  - pytest.ini and `[tool.pytest.*]` in pyproject.toml
  - etc.
- Use hatchling for building a new release only
- Update docs based on #118 
- Document path to v1.0 release?

Issue #118, opened 2025-07-11
Heavier Ax integration & a shift in focus


thomashopkins32:
Using https://ax.dev/ as a backend for Blop is a step in the right direction, in my opinion.

DOFs map to Parameters
Objectives map to Metrics (or rather a single OptimizationConfig)

There is a significant amount of overlap between Ax and the current state of Blop. I think if we focus on specifics involving beamline optimization (rather than black box optimization in general), we can move this project along faster.

So at a high level, here is my proposal:
- Wrap Ax Parameters using Blop DOFs (a DOF "has-a" single Parameter)
- Unpack Objectives into an Ax OptimizationConfig (for multi-objective there can be constraints)
- Blop Agent handles getting the next suggested points by the Ax framework, deploying them via Bluesky, and getting the result back via Tiled/Databroker and feeding those results to Ax for logging and further analysis.
  - Integration into bluesky-adaptive should be straightforward as well
- Ax handles experiment tracking, logging, visualization, optimization, constraints, & more
- Ax is flexible so we can continually expand access to custom models
- Blop switches to focusing on making beamline optimization more practical with:
  - Human-in-the-loop control
  - Collaborative optimization where beamline staff can suggest data points to try and this data is integrated into the models
  - Very safe optimization so the beam is not lost and equipment is not damaged
  - Specialized algorithms for beamline specific optimization problems

@NSLS-II/blop-team @jessica-moylan  What do you guys think?

PR #146, opened 2025-09-11
Deprecating older APIs

thopkins32:
# Summary
For #140, we are deprecating a bunch of the older API in favor of a transition toward more Ax friendly APIs. Since Ax will soon become the optimization engine backend for Blop (see #118), we should aim to simplify our APIs so that they can be easier to translate to Ax.

This cleanup of the APIs will make newer features much easier to implement and enable a stable, testable release cycle.

# Changes
Here is a high-level overview of the APIs we deprecate in this PR and why.

## agent.py
- `blop.agent.Agent` is no longer necessary now that `blop.ax.Agent` exists, they have (mostly) the same functionality, but the latter is much simpler to extend and maintain.
- We may want to revisit some of the things that `Agent` does that a standard `blop.ax.Agent` does not do currently. These include:
  - TSP solving for optimized measurement path (should be abstracted away as a blop-supported bluesky plan, in my opinion)
  - Validity conjugate modeling for constraint violations (could be implemented at the GPyTorch or BoTorch level, or as a custom Ax `GenerationStrategy`)
  - Multitask learning as a way to generalize across objectives (once again, either GPyTorch, BoTorch, or `GenerationStrategy`)

For the more complex modeling that the `blop.agent.Agent` does by default, I am not convinced that this should be default behavior. We need to analyze the various tradeoffs involved. The `blop.ax.Agent` provides a way to have custom models (and full generation strategies) so this should cover most, if not all complex use-cases that users of this software may have. See [this unit test](https://github.com/NSLS-II/blop/blob/76bd80510ca252924f1140c8a974fc3c6d4b8a8f/src/blop/tests/integrations/test_ax_agent.py#L81) for how it is done in practice.
## dofs.py
- No need for `DOFList` now that we don't manage DOFs ourselves (Ax does this for us)
- No more dependence on ophyd (use bluesky protocols instead)
- No more plotting related metadata, this should be tracked separately via Ophyd/Bluesky
- No more active/inactive DOFs (there are still read-only DOFs) due to the complexity in tracking, re-training, etc.

## objectives.py
- No need for `ObjectiveList` for similar reasons as above
- No need for plotting metadata
- Some arguments are better handled in digestion functions like:
  - transform
  - weight
  - min_noise
  - max_noise
- No more summaries for individual objectives (use `blop.ax.Agent.summary` instead)
- Models are no longer built and stored per-objective (there are Ax APIs to see this)


jennmald (approved):

jennmald (approved):
looks good

PR #154, opened 2025-09-29
Executable tutorials using jupyterlab

jessica-moylan:
Based on [executable tutorials](https://github.com/scientific-python/executable-tutorials) on Github, this makes the tutorials cleaner to commit while still allowing the ability to run them either through jupyter lab or myBinder.

In order to run the tutorials you can run ```pixi run run_tutorials``` which will open a jupyter lab session. These markdown-formatted tutorials can also be tested using ```pixi run local_test``` that temporarily converts the markdown files to .ipynb for testing purposes.


thopkins32:
Didn't have a chance to review fully yet, but I am curious. Why must we change the Python notebooks to markdown files? I'm not opposed to this, but I would like to understand why.

jessica-moylan:
When working on PR #143 , I realized that everytime I changed a tutorial file, I would also have the metadata that showed that something changed, which added additional information to the commits that weren't entirely neccessary. Additionally when reviewing changes to the `.ipynb` files, it was a bit difficult to see what changed.

Also, a few other repos use markdowns for their tutorials, tiled (uses .md) and bluesky (uses .rst), so it would make sense to follow in their example.

The big difference between this implimentation and those is that you are still able to execute the tutorials (either via juptiter lab or converting it back to the .ipynb file type)

thopkins32 (approved):

Issue #110, opened 2025-03-18
Add feature for blop to create JSON requests sent to the qserver rest api


SeijDeLeon:
Blop currently utilizes its own run engine for executing plans. If a beamline wants to only run plans through the Qserver then integrating blop with its full functionality requires going against this approach.

If blop could create a JSON request (eventually sent to the qserver through its REST API) then we could still run all plans through the qserver while keeping the functionality of the blop agent. Blop in this mode would not directly execute the plan through a run engine.

This functionality would readily support future interfaces which are centered around the queue server.

Sample diagram of proposed functionality:

![Image](https://github.com/user-attachments/assets/33253b8d-e22d-4444-9c60-efca113ff4c5)

@mrakitin @maffettone @whs92

maffettone:
This is what we built [bluesky-adaptive](https://github.com/bluesky/bluesky-adaptive) for.

I believe we made some changes to make this compatible in this PR: https://github.com/NSLS-II/blop/pull/86

whs92:
@maffettone are there any elements that blop has that adaptive doesn't have? I am new to both and have no preference

maffettone:
Good question. 

Bluesky adaptive is solely a harness, moving toward an API spec. While it has some example agents (basic bayesopt or sklearn stuff) It is meant to be agnostic to the kinds of logic you are using for your agent, aka bring-your-own agent. So in my view of this, BLOP brings the logic, bluesky-adaptive brings the integration harness. 

The design choice to integrate BLOP tightly with an `RE` or set of ophyd objects puts constrains on how it can be used. Which is fine, because the goal is to meet agent devs where they are at; however, hopefully #86 provides [at least the start of] an operating mode where the logic of BLOP is fully separate from the experimental orchestration.

maffettone:
Ready to try out! I just haven't had the time to write a good tutorial using bluesky pods. Because the async approach requires a bit more stack (tiled, message bus, qserver, etc).

Hopefully this is a good jumping off point. https://blueskyproject.io/bluesky-adaptive/how-to/extend-my-tool.html

We're hiring to get more active development on this, so feel free to be a squeaky wheel where this doesn't meet your needs

whs92:
I'll give it a go. We already have all of the rest of the stack in containers, it's easy to deploy. If I have questions I'll get back to you, better if I end up with a demo I'll try and share it.

whs92:
[...]

PR #212, opened 2025-12-14
Add support for Queueserver Agent

whs92:
In this merge request I try to address #110 by:

- Adding an agent which
    - is based on the Ax Agent class but that emits JSON to a Queueserver rather than messages to a RE
    - Is connected to a Tiled Server through the `EvalutationFunction`
    - Observes documents through a ZMQ buffer to work out if the requested plans have finished
- Adding a tutorial document which describes:
    - Changes required to a Queueserver Environment
    - Running plans remotely and observing the results using the agent

What's missing, is a link to a set of containers to start the required services. I think for people already using the queueserver the instructions will probably be sufficient, but I am interested to hear what others think.

I have tried to keep to the structure set out in v0.9.0 of blop where possible.

### DIFF
--- a/src/blop/ax/qserver_agent.py
+++ b/src/blop/ax/qserver_agent.py
@@ -3,10 +3,10 @@ import logging
 ###
 import threading
 import uuid
-from collections.abc import Callable, Sequence
-from typing import Any, Concatenate, ParamSpec
+from collections.abc import Sequence
+from typing import Any
 
-from ax.api.types import TOutcome, TParameterization
+from ax.api.types import TParameterization
 from bluesky.callbacks import CallbackBase
 from bluesky.callbacks.zmq import RemoteDispatcher
 from bluesky_queueserver_api import BPlan
@@ -65,8 +65,6 @@ class ZMQConsumer:
 
         self._zmq_thread = threading.Thread(target=self.zmq_consumer.start, name="zmq-consumer", daemon=True)
         self._zmq_thread.start()
-
-
 
 
 class BlopQserverAgent(BlopAxAgent):


docs/source/tutorials/qserver-experiment.md
```diff
@@ -0,0 +1,357 @@
...
+    jupytext_version: 1.18.1
+kernelspec:
+  display_name: .venv
+  language: python
+  name: python3
+---
+
+# Himmelblau Queueserver Demo
+
+In this demo we will make the himmelblau example, but with ophyd devices inside a remote qserer. The queueserver allows us to run the agent in another process, seperated from the experiment environment. The devices in the qserver include:
```
thopkins32:
```suggestion
In this demo we will make the himmelblau example, but with ophyd devices inside a remote qserver. The queueserver allows us to run the agent in another process, seperated from the experiment environment. The devices in the qserver include:
```

src/blop/ax/qserver_agent.py
```diff
@@ -0,0 +1,328 @@
...
+        # Should new suggestions be made automatically until all of the trials are complete?
+        self.continuous_suggestion = True
+
+        # Learning parameters
+        self.num_itterations = 30
+        self.n_points = 1
+
+        # Variables used to keep track of the current optimization
+        self.current_itteration = 0
+        self.agent_suggestion_uid = None
```
thopkins32: Is it possible to get the BlueskyRun `uid` from the plan that is run by the queueserver? That way you won't have to search the Tiled Catalog for this custom `agent_suggestion_uid` from the metadata but can index the catalog directly.

whs92: I think unfortunately it's not possible yet. In part because you might not execute a plan immediately. You would have to subscribe to changes and be told when a plan was running and what's it's uid is. @tacaswell do you know otherwise? 

I guess another approach would be if the uid can be specified in the md dict, but I guess then it wouldn't be guaranteed to be unique. 

Adding my own key and then searching for it seemed like the easiest solution. Also, when I used this in a real application I needed to submit multiple plans then keep track of them and reconstruct data that was acquired. Adding various keys was helpful for this.

src/blop/ax/qserver_agent.py
```diff
@@ -0,0 +1,326 @@
...
+        If it has completed, we can digest the data from it and move on to the next point.
+        """
+
+        if self._listen_to_events:
+            # Mark the current acquisition as finished
+
+            logger.info("A stop document has been received, evaluating")
+
+            # Evaluate it with the evaluation function
+            outcomes = self.optimization_problem.evaluation_function(uid=self.agent_suggestion_uid, suggestions=self.trials)
```
thopkins32: `start_doc` should have a `uid` if you prefer to use that over searching. And this is called when the stop doc is generated so it should exist in the database at this point.

thopkins32 (approved):
LGTM. Thanks @whs92 !

Issue #190, opened 2025-11-25
Documentation planning


thopkins32:
Here are some ideas for improving the documentation:
- bluesky-adaptive example integration
- interchangeable optimizer backends (Xopt, etc.)
- ophyd and ophyd-async examples
- tiled and databroker examples
- flyscanning (still need to update protocols for this I think)

PR #215, opened 2025-12-18
Tiled databroker how-to guide

jessica-moylan:
Part of #190

### DIFF
--- a/docs/source/how-to-guides/tiled-databroker.rst
+++ b/docs/source/how-to-guides/tiled-databroker.rst
@@ -90,16 +90,9 @@ To access the data for optimization, you have to connect to a Tiled server or Da
 
 .. testcode::
 
-    from blop.sim import HDF5Handler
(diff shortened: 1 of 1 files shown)


thopkins32 (changes requested):
Looks good! It would be great if you can make it testable like the other how-to-guides.

If you make it use `.. testsetup::`, `.. testcode::`, and optionally `.. testoutput`. Then we can test the code in the document using
```
pixi run doc-tests
```

I typically setup all of the assumptions and mock things in a hidden `.. testsetup::` section at the top of the file, and then have the rest use that.

In some of them I even spin up a local tiled server: https://github.com/NSLS-II/blop/blob/8ee29a34d9e984c51e3f9694b7292b131d73e6c7/docs/source/how-to-guides/acquire-baseline.rst?plain=1#L68-L73

After that it should be good to merge!

PR #252, opened 2026-02-26
Add support for `fixed_parameters`

### LINKED ISSUE #244: Fixed parameters support
https://ax.readthedocs.io/en/stable/api.html#ax.api.client.Client.get_next_trials has a `fixed_parameters` argument that is currently not supported.

From my understanding, this allows you to set a partial parameterization (fixing the values of some parameters and making suggestions with the rest), while continuing to track all of them.

josephhanrahan:
Users can now indicate DOFs (parameters in the Ax context) to the agent to be fixed across `suggest` calls.

Closes #244

### DIFF
--- a/src/blop/ax/agent.py
+++ b/src/blop/ax/agent.py
@@ -197,7 +197,9 @@ class Agent:
         elif _has_dof_keys(fixed_dofs):
             self._optimizer.fixed_parameters = {dof.parameter_name: value for dof, value in fixed_dofs.items()}
         else:
-            raise ValueError(f"Keys must all be either {type(DOF)} or {type(str)}, but got {type(list(fixed_dofs.keys())[0])}")
+            raise ValueError(
+                f"Keys must all be either {type(DOF)} or {type(str)}, but got {type(list(fixed_dofs.keys())[0])}"
+            )
 
     def to_optimization_problem(self) -> OptimizationProblem:
         """


src/blop/ax/agent.py
```diff
@@ -49,6 +49,8 @@ class Agent:
         :func:`blop.plans.default_acquire` will be used.
     dof_constraints : Sequence[DOFConstraint] | None, optional
         Constraints on DOFs to refine the search space.
+    fixed_dofs : dict[str, Any] | None, optional
+        A mapping of DOF names to the values they should be fixed to.
```
thopkins32: Don't think this should be set on `__init__` since its a fairly special case.
```suggestion
```

src/blop/ax/agent.py
```diff
@@ -81,6 +83,7 @@ def __init__(
         evaluation_function: EvaluationFunction,
         acquisition_plan: AcquisitionPlan | None = None,
         dof_constraints: Sequence[DOFConstraint] | None = None,
+        fixed_dofs: dict[str, Any] | None = None,
```
thopkins32: No need in `__init__`
```suggestion
```

src/blop/ax/agent.py
```diff
@@ -165,6 +169,14 @@ def ax_client(self) -> Client:
     def checkpoint_path(self) -> str | None:
         return self._optimizer.checkpoint_path
 
+    @property
+    def fixed_dofs(self) -> dict[str, Any] | None:
+        return self._optimizer.fixed_parameters
+
+    @fixed_dofs.setter
+    def fixed_dofs(self, fixed_dofs: dict[str, Any] | None) -> None:
+        self._optimizer.fixed_parameters = fixed_dofs
```
thopkins32: Include the ability to use DOFs as keys.
```suggestion
    def fixed_dofs(self, fixed_dofs: dict[DOF, Any] | dict[str, Any] | None) -> None:
        if all(isinstance(key, DOF) for key in fixed_dofs.keys()):
            self._optimizer.fixed_parameters = { dof.parameter_name: value for dof, value in fixed_dofs }
        elif all(isinstance(key, str) for key in fixed_dofs.keys()):
            self._optimizer.fixed_parameters = fixed_dofs.copy()
        else:
            raise ValueError(f"Keys must all be either {type(DOF)} or {type(str)}")
```

src/blop/ax/optimizer.py
```diff
@@ -89,6 +94,20 @@ def checkpoint_path(self) -> str | None:
...
+        return self._fixed_parameters
+
+    @fixed_parameters.setter
+    def fixed_parameters(self, fixed_parameters: dict[str, Any] | None) -> None:
+        if fixed_parameters is None:
+            self._fixed_parameters = None
+            return
+        unknown_parameter_names = set(fixed_parameters) - set(self._parameter_names)
+        if unknown_parameter_names:
+            raise KeyError(f"Unknown fixed parameter(s): {sorted(unknown_parameter_names)}")
```
thopkins32:
```suggestion
            raise KeyError(f"Unknown fixed parameter(s): {sorted(unknown_parameter_names)}, expected: {sorted(self._parameter_names)}")
```

PR #268, opened 2026-03-13
Reconfigurable search spaces

### LINKED ISSUE #245: Reconfigurable search spaces
Ax's Client API doesn't support reconfiguring the search space. You must create a new Client. This is an issue because we likely want to carry over the data from the previous search space.

So a feature that enables moving data from one Client to another would be useful to have. Basically you have to re-attach trials from one client to the next. Probably something that should happen at Blop's Agent-level rather than lower down in the optimizer.

josephhanrahan:
Allow updating `RangeDOF` bounds mid-experiment with `Agent.reconfigure_search_space()`.

Closes #245


thopkins32 (commented):
Discussed offline:
- `DOF` key handled at agent level
- `str` key handled at optimizer level
- `ChoiceDOF` support

src/blop/ax/optimizer.py
```diff
@@ -214,3 +216,91 @@ def checkpoint(self) -> None:
...
+        for parameter_name, value in original_range_values.items():
+            parameter = self._client._experiment.parameters[parameter_name]
+            if isinstance(parameter, RangeParameter):
+                parameter.update_range(*value)
+        for parameter_name, value in original_choice_values.items():
+            parameter = self._client._experiment.parameters[parameter_name]
+            if isinstance(parameter, ChoiceParameter):
+                parameter.set_values(value)
+
+    def _reconfigure_search_space(
```
thopkins32: We can probably make this public
```suggestion
    def reconfigure_search_space(
```

src/blop/ax/optimizer.py
```diff
@@ -214,3 +216,91 @@ def checkpoint(self) -> None:
...
+        TypeError
+            If the provided value does not match the expected type for the parameter.
+        """
+        parameter = self._client._experiment.parameters[parameter_name]
+        if isinstance(parameter, RangeParameter):
+            if isinstance(value, tuple) and len(value) == 2 and all(isinstance(x, float | int) for x in value):
+                original_range_values[parameter_name] = (parameter.lower, parameter.upper)
+                parameter.update_range(*value)
+            else:
+                raise TypeError(f"Expected range to be a tuple of two floats, but got {value}")
```
thopkins32: I wonder if Ax raises the TypeError for us in `parameter.update_range(*value)`. We may not have to do an extra `all(isinstance(x, ...) for x in value)` check.

Same thing for `set_values` below.

src/blop/ax/optimizer.py
```diff
@@ -214,3 +216,91 @@ def checkpoint(self) -> None:
...
+        """
+        Update the bounds or values of existing parameters in the underlying experiment
+
+        Parameters
+        ----------
+        parameter_mappings : dict[str, tuple[float, float] | list[float] | list[int] | list[str] | list[bool]]
+            Mapping of parameter names to (lower, upper) bounds or a list of values depending on the parameter type.
+
+        """
+        self._verify_parameter_names(set(parameter_mappings))
```
thopkins32: I wonder what the consequence of not validating is. What does Ax say when we give it a parameter that doesn't exist?

thopkins32 (commented):
Looks correct. I wonder if we can reduce the amount of validation we do. I imagine Ax does some internal validation. They have more parameter metadata so they can probably validate much better than we can.

Can you check what happens with your current tests if we remove all validation logic?

thopkins32 (approved):

PR #272, opened 2026-04-07
Failed and abandoned suggestions

### LINKED ISSUE #261: Handle failed and abandoned suggestions
Failed Bluesky runs need to be marked as failed. Interrupted Bluesky runs should be marked as abandoned. It would be best if we can try to salvage data from the run if `num_points > 1` per optimization step.

We should also provide a way to manually mark previous trials as abandoned or failed, regardless of their current state.

MTakahashi-KWH:
initial implementation to resolve #261 . Will add abandoned casing once bluesky run engine interrupt sequence is better understood (need to capture some types of exceptions: keyboardinterr,....).

### DIFF
--- a/src/blop/ax/optimizer.py
+++ b/src/blop/ax/optimizer.py
@@ -5,10 +5,10 @@ from ax import ChoiceParameterConfig, Client, RangeParameterConfig
 from ax.core.parameter import ChoiceParameter, RangeParameter
 from ax.core.types import TParamValue
 
-from ..protocols import ID_KEY, CanRegisterSuggestions, Checkpointable, Optimizer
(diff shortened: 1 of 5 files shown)


thopkins32 (commented):
Looks good so far.

The typical abort flow for the RunEngine is:
```
2x SIGINT -> RE.state == "paused" -> RE.abort()
```
but I am uncertain if we can know that the RE aborted within a plan. If we can reliably do so, then we should mark all trials that haven't been run as abandoned. Otherwise, they may have to manually mark them that way. In this case we likely don't need a special protocol for that, we should merely document how to do it with the Ax client.

What do you think?

Add me back for a full review when the CI checks pass and the PR is out of draft.

thopkins32 (approved):

PR #280, opened 2026-04-20
added Graphviz to _docs.yml to fix ExecutableNotFound error

jessica-moylan:
Fixes `ExecutableNotFound: failed to execute PosixPath('dot'), make sure the Graphviz executables are on your systems' PATH` on tutorials in the published docs

### DIFF
--- a/.github/workflows/_docs.yml
+++ b/.github/workflows/_docs.yml
@@ -30,6 +30,11 @@ jobs:
           environments: docs
           cache: false
           activate-environment: docs
+      
+      - uses: actions/checkout@v4
+      - name: Setup Graphviz
+        uses: ts-graphviz/setup-graphviz@v2
+
 
       - name: Test with doctest
         run: pixi run doc-tests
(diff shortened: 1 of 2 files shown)


thopkins32:
@jessica-moylan Can you provide more information about where this error is coming from?

jessica-moylan:
@thopkins32 one such example can be found here [http://blueskyproject.io/blop/main/tutorials/xrt-kb-mirrors.html#checking-optimization-health](http://blueskyproject.io/blop/main/tutorials/xrt-kb-mirrors.html#checking-optimization-health) but can be found in other locations when graphs are trying to be displayed. How this fixes the problem can be seen in the github action artifacts

thopkins32:
@jessica-moylan I see now, can you add `graphviz` as a conda dependency in the Pixi docs environment instead of directly in the CI workflow? I believe it should work.

It looks like the PyPI version of graphviz only installs the Python bindings, not the underlying executable. The executable is available from conda-forge.

thopkins32 (approved):

PR #291, opened 2026-05-05
Fix stop_listener and incorrect usage of UID in stop document comparison in queueserver

### LINKED ISSUE #289: BUG: incorrect comparison of UIDs in the queueserver runner callback
The stop doc comparison uses doc["uid"] which is the stop document's own unique ID, not the run it belongs to. It should use doc["run_start"] which references the corresponding start document's UID.

### LINKED ISSUE #290: BUG: queueserver stop_listener called from within its own event loop
After the final iteration, `_on_acquisition_complete` calls `self._client.stop_listener()` which calls `self._dispatcher.stop()` which tries to close the asyncio event loop. But this callback is executing inside that same event loop (on the qserver-zmq-consumer thread), causing `RuntimeError: Cannot close a running event loop.`

thopkins32:
Closes #290 
Closes #289 

Also, adds some missing properties and the `stop` method, both of which are useful to inspect the state of the runner. This is because `agent.run()` returns immediately and the optimization loop runs in a background thread.

### DIFF
--- a/src/blop/ax/agent.py
+++ b/src/blop/ax/agent.py
@@ -603,6 +603,17 @@ class QueueserverAgent(_AxAgentMixin):
     def acquisition_plan(self) -> str | None:
         return self._acquisition_plan
 
+    @property
+    def is_running(self) -> bool:
+        return self._runner.is_running
+
+    @property
+    def current_iteration(self) -> int:
+        return self._runner.current_iteration
+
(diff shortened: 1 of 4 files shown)


MTakahashi-KWH (approved):
looks good. Direct bug fix to allow queueserver optimization to close out correctly

PR #292, opened 2026-05-05
Bluesky Queueserver tutorial

### LINKED ISSUE #213: Create instances of Queueserver, Tiled and ZMQ proxy that can be run during tutorials and tests
In order to allow people to easily run the tutorials with the blop queueserver agent, we need to be able to create a queueserver, tiled and zmq instance. At the moment I am doing this in containers with docker compose. I could make this available, but maybe there is an easier way?

thopkins32:
- Adds tutorial for how one might set up a distributed architecture for using Blop with Bluesky Queueserver & Tiled, repeating the himmelblau example (since the optimization part isn't the focus of the tutorial)
- The tutorial depends on external services, so I created the minimal necessary containers to be run before the tutorial starts
- I simplified the pixi tasks related to building the docs, so that everything spins up and shuts down properly

Closes #213 

I'm open to ideas for improving this architecture...

### DIFF
--- a/.github/workflows/_docs.yml
+++ b/.github/workflows/_docs.yml
@@ -34,15 +34,6 @@ jobs:
       - name: Start queueserver tutorial infrastructure
         run: |
           docker compose -f docs/source/tutorials/queueserver/docker-compose.yml up -d --build --wait
-          # Open the RE Manager worker environment
-          pixi run python -c "
-          from bluesky_queueserver_api.zmq import REManagerAPI
-          RM = REManagerAPI(zmq_control_addr='tcp://localhost:60615')
-          RM.environment_open()
-          RM.wait_for_idle(timeout=60)
-          print(RM.status())
-          RM.close()
-          "
 
       - name: Test with doctest
         run: pixi run doc-tests
--- a/.github/workflows/_docs.yml
+++ b/.github/workflows/_docs.yml
@@ -31,19 +31,8 @@ jobs:
           cache: false
           activate-environment: docs
 
-      - name: Start queueserver tutorial infrastructure
-        run: |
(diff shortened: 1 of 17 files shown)


thopkins32:
Download and inspect the docs here: https://github.com/bluesky/blop/actions/runs/25446447128/artifacts/6835336895

MTakahashi-KWH (approved):
I approve the code as is, but some things of note:
- the demo doesn't actually show any convergence of the queueserver agent, just random point sampling essentially. Maybe show some information about best point guess or allow more iterations to converge at some point.
- The fact that the evaluation function now has to take in account the intrinsics of the queueserver as well as the data management is of major concern. We are shifting too much responsibility with EvaluationFunction turning into a god function, essentially expecting users to "rebuild a pipeline" out to our interface. I suggest we look into structuring regarding separating data handling and processing, or providing data handling interfaces.

thopkins32:
@MTakahashi-KWH thanks for reviewing
> the demo doesn't actually show any convergence of the queueserver agent, just random point sampling essentially. Maybe show some information about best point guess or allow more iterations to converge at some point.

This is intentional. The point of the tutorial is the architecture, not the optimization. There are other tutorials for that.

> The fact that the evaluation function now has to take in account the intrinsics of the queueserver as well as the data management is of major concern. We are shifting too much responsibility with EvaluationFunction turning into a god function, essentially expecting users to "rebuild a pipeline" out to our interface. I suggest we look into structuring regarding separating data handling and processing, or providing data handling interfaces.

I'm open to ideas to improve this. However, I partially disagree with "separating data handling and processing". I'm skeptical we can provide a general-purpose data handling abstraction that doesn't become limiting. For example, users may need to load large datasets from custom file stores or move data between CPU and GPU, which are tightly coupled to their processing logic.

PR #293, opened 2026-05-06
Best point(s) navigation

### LINKED ISSUE #243: Best point(s) navigation
Make it easier re-run the best trial or run the best trial against a different Bluesky plan (not necessarily the configured acquisition plan).

Difficulty here is how to handle multiple best-points (i.e. Pareto frontier)

thopkins32:
Closes #243

### DIFF
--- a/src/blop/ax/agent.py
+++ b/src/blop/ax/agent.py
@@ -131,11 +131,11 @@ class _AxAgentMixin:
 
         Returns
         -------
-        list[tuple[str, dict, dict]]
+        list[tuple[int, TParameterization, TOutcome]]
             Each element in the list is a tuple of:
-              - "_id" of the suggestion
-              - suggested parameters
-              - measured outcomes
+              - trial index (int)
+              - parameter values (dict)
+              - metric values (dict, where values may be (value, sem) tuples)
 
         See Also
         --------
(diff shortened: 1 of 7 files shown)


MTakahashi-KWH:
Would best points be a better protocol option within optimizer itself instead of as another interface? 
I'd think that its pretty implicit that optimizer needs to state what it found as optimal to be able to close off its optimization. I think shifting the protocol for optimizer to have a "get_optimum" function instead of the "HasBestPoints" interface is best, and navigate_to_best can easily refactor for this, given they are essentially inseparable. 

if for some absolute reason that an optimizer can't directly report its optimal, a bypass cache which just caches the sampled point which the evaluation function reports as lowest is easily implementable within an optimizer's ingest structure.

Expansive optimum reports like trust region solutions and solution subspace would be the extending interface if need be.

thopkins32:
@MTakahashi-KWH agreed, good point. My instinct was to not touch existing interfaces but I think this is reasonable. I'll refactor with that in mind.

MTakahashi-KWH (approved):
Nice, looks good to me. Enables for parametrization in the future for optimize() plans to either stop at start position or the optimal, instead of last sampled point.

PR #296, opened 2026-05-08
Prevent parallel runs of the queueserver runner

thopkins32:
Two threads would be modifying the same optimization state, which we can't allow.

Locking is necessary to synchronize read/write of the optimization state between the main thread and the ZMQ dispatcher thread's callback.


thopkins32:
Realizing that this needs a larger refactor...going to do that now.

MTakahashi-KWH (approved):
looks good. Internal locks to allow atomics/criticals so no intermediary state is reported to the zmq proxy

MTakahashi-KWH:
eh, would also note deletions of submit next and submit next manual. Ostensibly, those functions aren't necessary anymore, coming from a prompted update to a full plan submitted.

PR #301, opened 2026-05-13
Fix ChoiceDOF Bug

### LINKED ISSUE #299: ChoiceDOF Fixed Parameter Bug
`ChoiceDOF` cannot be used as a key in the `Agent.fixed_dofs` dictionary because lists as attributes of a dataclass aren't hashable.

The first trial also always defaults to CenterOfSearchSpace, which overrides any fixed parameters. We should decide whether or not we want this behavior.

josephhanrahan:
Since both `Agent.fixed_dofs` and `Agent.reconfigure_search_space()` accept type `dict[ DOF, ...]` as input, ChoiceDOF can't be used as a dictionary key because it isn't hashable.

This change implements `ChoiceDOF.__hash__()`, having it return `hash(self.parameter_name)`, which is safe because Ax raises a ValueError if parameter names aren't unique (see `test_agent_init_duplicate_dof_names`).

Also added test coverage for using ChoiceDOF with `fixed_dofs` and `reconfigure_search_space()`.

Closes #299

### DIFF
--- a/src/blop/ax/dof.py
+++ b/src/blop/ax/dof.py
@@ -174,6 +174,9 @@ class ChoiceDOF(DOF):
             dependent_parameters=self.dependent_parameters,
         )
 
+    def __hash__(self):
+        return hash(self.parameter_name)
+
 
 class DOFConstraint:
     """
--- a/src/blop/ax/dof.py
+++ b/src/blop/ax/dof.py
@@ -62,7 +62,7 @@ class DOF(ABC):
     def to_ax_parameter_config(self) -> RangeParameterConfig | ChoiceParameterConfig: ...
 
 
-@dataclass(frozen=True, kw_only=True)
+@dataclass(frozen=True, kw_only=True, eq=False)
 class RangeDOF(DOF):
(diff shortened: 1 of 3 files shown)


src/blop/ax/dof.py
```diff
@@ -174,6 +174,9 @@ def to_ax_parameter_config(self) -> ChoiceParameterConfig:
             dependent_parameters=self.dependent_parameters,
         )
 
+    def __hash__(self):
+        return hash(self.parameter_name)
```
thopkins32: Maybe this belongs on the base `DOF` class instead?

josephhanrahan:
Using `frozen=True` and `eq=False` together with `@dataclass` defaults to using the `__hash__` and `__eq__` of the superclass, in this case `DOF`, which will then compare and hash using `self.name` or `self.actuator`. This gives the same behavior as what I originally proposed. (https://docs.python.org/3/library/dataclasses.html#module-contents)

thopkins32 (approved):

PR #303, opened 2026-05-13
simple tutorial: suppress Ax dep warning, fixed reorder in evaluation, better convergence

MTakahashi-KWH:
Suppressed the spammy "groupby" warning from Ax usage of pandas (not our code) so you don't have to scroll the output log. Added point reordering correction to account for shortest path optimization done by the acquisition plan. Switched the optimization to run 5 iterations of 8 points just to get cleaner coverage of the space and better convergence (at least show it works)

### DIFF
--- a/docs/source/tutorials/simple-experiment.md
+++ b/docs/source/tutorials/simple-experiment.md
@@ -23,6 +23,7 @@ First, let's import what we need and start the data infrastructure:
 import logging
 import time
 from typing import Any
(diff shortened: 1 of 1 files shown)


thopkins32 (approved):
Nicely done. I like the print message showing the reordering. Agreed on increasing the num points per iteration to showcase that, otherwise it's unclear what the point of that is.

PR #305, opened 2026-05-15
Use futures for better tracking of queueserver runner state

### LINKED ISSUE #304: Background thread exceptions on queueserver runner
For when things fail, the state needs to be updated appropriately

thopkins32:
Yet another refactor of the queuserver integration...

Closes #304 

# Summary

Tracking the optimization progress was annoying because you had to poll `agent.is_running`. There was also zero failure handling so when things went wrong it was impossible to tell and it looks like the run just got stuck somewhere.

Now we are properly handling the asynchronous nature of the queueserver integration by using `concurrent.futures`. We set exceptions when they occur. We have something that can block until the run is complete (no more polling!). And it now atuomatically registers failures with the optimizers than can handle them.

I also update the tutorial to use the future instead of the polling loop. I add to the docs reference all of the queueserver runner stuff as well.

### DIFF
--- a/docs/source/reference.rst
+++ b/docs/source/reference.rst
@@ -11,3 +11,4 @@ API Reference
     reference/outcome-constraint.rst
     reference/plans.rst
     reference/protocols.rst
+    reference/queueserver.rst
--- a/docs/source/reference/queueserver.rst
+++ b/docs/source/reference/queueserver.rst
@@ -1,0 +1,29 @@
+Queueserver
+===========
+
+These classes implement the distributed optimization backend that connects
+Blop to a remote `Bluesky Queueserver <https://blueskyproject.io/bluesky-queueserver/>`_.
+See the :doc:`/tutorials/queueserver` tutorial for a full worked example.
+
+OptimizationResult
+------------------
+
+.. autoclass:: blop.queueserver.OptimizationResult
+   :members:
+   :undoc-members:
+
+QueueserverClient
+-----------------
+
+.. autoclass:: blop.queueserver.QueueserverClient
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
+QueueserverOptimizationRunner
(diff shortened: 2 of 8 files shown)


MTakahashi-KWH (approved):
straightforward refactoring to shift from a queueserver polling model to a more classic thread sync and event model. major refactoring of test cases to accept this model, with names like: "test_runner_on_acquisition_complete_uid_mismatch_sets_future_exception"

 note: final initialization of the zmq dispatch has no validation within _start_listener thread due lack of monitoring support on the remote dispatcher. this is noted by the TODO marker in the code and is an upstream concern. Current implementation should be fail safe by marking future as failed when submitting suggestion inevitably fails.

extreme side note: Any interest in publishing optimize_step or optimize to a queue server to make a queue server agent implementation act like a regular agent in the document stream (s.t. no double registering/ tiled writing is similar)? Current abstraction already puts the agent object independent of the optimization plan. May be a "queueserver needs document/reference injection" as a plan more than anything.

thopkins32:
> extreme side note: Any interest in publishing optimize_step or optimize to a queue server to make a queue server agent implementation act like a regular agent in the document stream (s.t. no double registering/ tiled writing is similar)? Current abstraction already puts the agent object independent of the optimization plan. May be a "queueserver needs document/reference injection" as a plan more than anything.

Not sure if this is appropriate. My intended usage of an optimizer + queueserver is to have the optimizer act as a queueserver client the way a normal operator would. The optimizer (or maybe the runner) is responsible for tracking and reporting optimization state, something that I have not got around to just yet. This PR is a meaningful step toward that since it now reports errors. It sort of is available if the optimizer implements it like the Ax client does with `summarize()` and the various analyses. And since the agent submitting iterations is done in the background thread, you have free access to these analyses while data is being acquired.

PR #306, opened 2026-05-18
Add plan keyword arguments to queueserver optimization problems

thopkins32:
Small oversight on how remote execution differs from in-process. In-process (with the `RE`) can always add additional arguments and wrap with `functools.partial`, but this can't happen through remote execution with queueserver. Therefore, we need to allow for keyword arguments for plans.

I also took this opportunity to clean up some related docstrings.

### DIFF
--- a/src/blop/ax/agent.py
+++ b/src/blop/ax/agent.py
@@ -602,7 +602,7 @@ class QueueserverAgent(_AxAgentMixin):
     blop.ax.dof.ChoiceDOF : For discrete parameters.
     blop.ax.objective.Objective : For defining objectives.
     blop.ax.optimizer.AxOptimizer : The optimizer used internally.
-    blop.queueserver.QueueservverOptimizatonRunner : Runner that handles interaction with bluesky-queueserver.
+    blop.queueserver.QueueserverOptimizatonRunner : Runner that handles interaction with bluesky-queueserver.
     """
 
     def __init__(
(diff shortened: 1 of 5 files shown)


MTakahashi-KWH (commented):
- pull request adds kwargs passing in agents for acquisition plans. fixed servvver

QueueserverAgent now has 11 named initialization parameters, we should think about either pushing pass through values to kwargs, or (in the future) exposing an optimization config object for agents which doesn't chicken or egg the OptimizationProblem made by the agent.

thopkins32:
@MTakahashi-KWH only 6 are required arguments though. 11 is not so bad if you take a look at https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html#pandas.read_csv

Adding one level of indirection via some "config" object is not worth being part of the public API, in my opinion. We can move to kwargs if needed, although I don't think there will be significantly more than 11 arguments anytime soon.

MTakahashi-KWH (approved):
okay, approving though footnoting the lack of interaction with the agent itself as a named pass-through argument. A config object should also be backburnered with the structural idea of an agent's responsibility being "we provide the problem, the agent orchestrates it into an optimization program"

...Perhaps there is a space where the agent or optimizer is interested in tweaking underlying characteristic/ arguments of the acquisition plan during runtime which are not parameters themselves (aquisition times, recalibrations ... as evaluations fail/ become noisy ...)... which would require the evaluation function to throw its own trial faults (maybe those are marked abandoned).

thopkins32:
> ...Perhaps there is a space where the agent or optimizer is interested in tweaking underlying characteristic/ arguments of the acquisition plan during runtime which are not parameters themselves (aquisition times, recalibrations ... as evaluations fail/ become noisy ...)... which would require the evaluation function to throw its own trial faults (maybe those are marked abandoned).

Wouldn't this change the underlying data distribution? Changing an acquisition time mid-optimization, for example, changes the statistics of the detector readout.

PR #308, opened 2026-05-22
Replace ZMQ address argument with RemoteDispatcher for queueserver

thopkins32:
# Summary

Currently the `QueueserverAgent`, `QueueserverOptimizationRunner`, and the `QueueserverClient` class all accept a ZMQ address as a string and eventually the client constructs a `RemoteDispatcher` internally using this address. This is a poor design since it doesn't allow for a user to use custom dispatchers or pass additional arguments to `RemoteDispatcher` (such as a prefix).

This PR replaces the ZMQ address argument with a document dispatcher argument that enables using an instance of a RemoteDispatcher that has already been configured.

# Breaking changes

- API change for the agent, runner, and client classes listed above

### DIFF
--- a/src/blop/ax/agent.py
+++ b/src/blop/ax/agent.py
@@ -17,6 +17,7 @@ else:
 # ===============================
 import bluesky.preprocessors as bpp
 from bluesky.callbacks import CallbackBase
+from bluesky.callbacks.zmq import RemoteDispatcher
 from bluesky.utils import MsgGenerator
 from bluesky_queueserver_api.zmq import REManagerAPI
 
@@ -571,9 +572,8 @@ class QueueserverAgent(_AxAgentMixin):
     ----------
     re_manager_api : REManagerAPI
         The manager API for interaction with Bluesky queueserver.
-    zmq_consumer_addr : str
-        A ZMQ address to consume Bluesky messages from, to react to plan execution on the
-        remote server.
+    document_dispatcher : RemoteDispatcher
+        Dispatcher for consuming Bluesky documents from the remote server.
     sensors : Sequence[str]
         The sensors to use for acquisition. These should be the minimal set
         of sensors that are needed to compute the objectives.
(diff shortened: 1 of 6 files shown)


src/blop/ax/agent.py
```diff
@@ -608,7 +608,7 @@ class QueueserverAgent(_AxAgentMixin):
     def __init__(
         self,
         re_manager_api: REManagerAPI,
-        zmq_consumer_addr: str,
+        document_dispatcher: RemoteDispatcher,
```
jwlodek, line 611: Is there any case where you'd want an in-process dispatcher? Perhaps loosening the type hint to `DocumentDispatcher` would work? Or are there zmq specifics in the `QueueServerClient` here?

thopkins32: I thought about this and decided against it, for now.

Internally we handle the background listener thread with `dispatcher.start`. We also register our own callback with the dispatcher before starting it (not sure if that order matters). The base `Dispatcher` class doesn't have a `start` or a `stop` method. 

I'm also not sure we should delegate to the user to start the background thread properly if most cases are going to use `RemoteDispatcher` anyway. Technically we don't have any dependency on ZMQ here, someone could subclass `RemoteDispatcher` and overwrite all of the ZMQ parts. We need three methods: `start`, `stop`, and `subscribe` on the dispatcher.

thopkins32: If someone has a use-case where they must be responsible for the lifecycle of `dispatcher.start` and `dispatcher.stop`, then we can re-think this. But for ease-of-use, I think Blop can take care of it.

MTakahashi-KWH (approved):
Flip dispatcher to use Dependency Injection. Does reduce off the shelf capability of queue server agent and adds responsibility to the implementer, but, such an agent will probably be defined in a long term or nearly static scope so upfront implementation costs are acceptable.

PR #311, opened 2026-05-25
XRT KB mirror tutorial improvements

thopkins32:
# Summary

- Rather than optimizing three separate objectives, we minimize only one: FWHM
- We do 1 center of search space sample + 9 random samples
- We also track `intensity` and ensure that it is > 10,000 to avoid complete beam loss
- Replaced `"auto"` setting for z-axis toroid mirror, which could be considered "cheating", with a fixed value

## Results

This now consistently gets max intensity of 50,000 rays and finds the optimal parameters in fewer iterations.


josephhanrahan (approved):
Nice improvements. All looks good to me.

jessica-moylan:
I agree, looks good to me. I especially like the constraint to have a certain amount of beam.

jessica-moylan (approved):

Issue #295, opened 2026-05-07
Flaky Queueserver tutorial


thopkins32:
I believe it is related to a slow-joiner problem with ZMQ. Issue opened upstream: https://github.com/bluesky/bluesky/issues/2014

Also, a PR: https://github.com/bluesky/bluesky/pull/2015

whs92:
I never saw this in the past because the thread was opened once when the agent was initialized, I then used flags to decide if the callback should do anything if an event was received. I get that that is a bit clunky though.

I think we should rewrite this part completely now that Tiled Websockets is a thing. we can do away with any of the ZMQ stuff. Or, as I think your suggesting in other comments we do away with the zmq stuff entirely and just poll tiled, which is slower but more robust.

thopkins32:
It probably should be started on `__init__` which would make it less likely. The error has been very rare so far and really only somewhat likely on CI runners. The linked Bluesky PR should resolve it completely either way.

Agreed, Tiled Websockets is likely the path forward.

PR #316, opened 2026-06-02
Start queueserver listener in runner init

whs92:
Starts the queueserver document listener once during `QueueserverOptimizationRunner` initialization instead of before each plan submission. This ensures the listener is already active before queueserver runs can emit their first documents. It tried to address the problem in #295. This is a stop gap until we move to TiledWebSockets.

Tests were updated to verify:

- `start_listener()` is called once at runner construction.
- `run()` and `submit_suggestions()` do not restart the listener.
- Multi-iteration runs keep using the same listener.
- Listener startup failures now raise during construction, while `submit_plan` failures still fail the active future and re-raise.

### DIFF
--- a/src/blop/queueserver.py
+++ b/src/blop/queueserver.py
@@ -479,7 +479,7 @@ class QueueserverOptimizationRunner:
 
     def _on_acquisition_complete(self, start_doc: RunStart, stop_doc: RunStop) -> None:
         """Callback when acquisition finishes. Ingest results and maybe continue."""
-        
+
         try:
             self._process_acquisition(start_doc, stop_doc)
         except Exception as exc:
@@ -495,7 +495,6 @@ class QueueserverOptimizationRunner:
     def _process_acquisition(self, start_doc: RunStart, stop_doc: RunStop) -> None:
         """Core acquisition-complete logic (called from _on_acquisition_complete)."""
 
-        
         with self._state_lock:
             if self._state is None:
                 raise RuntimeError("_on_acquisition_complete called before run()")
--- a/src/blop/queueserver.py
+++ b/src/blop/queueserver.py
@@ -79,10 +79,17 @@ class ConsumerCallback(CallbackBase):
         super().__init__()
         self._start_doc_cache: RunStart | None = None
         self._callback = callback
+        self._start_doc_filter: Callable[[RunStart], bool] | None = None
+
+    def set_start_doc_filter(self, start_doc_filter: Callable[[RunStart], bool] | None) -> None:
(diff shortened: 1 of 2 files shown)


src/blop/queueserver.py
```diff
@@ -210,12 +219,15 @@ def start_listener(self, on_stop: Callable[[RunStart, RunStop], None]) -> None:
         on_stop : callable
             Callback invoked when a stop document is received.
             Signature: on_stop(start_doc, stop_doc)
+        start_doc_filter : callable, optional
+            Predicate used to ignore unrelated start documents before caching.
```
thopkins32: Curious why you may need a custom start doc filter. Is the correlation UID injection + filter not enough?

whs92: You are right. I will remove it

whs92:
I need to test once more with a real queueserver environment not just the mocks in the pytests

whs92:
I tested with the real queueserver, not just the tests in the pipeline. I removed the unused additional check. 

The docs are not building, but that isn't anything I have changed.

src/blop/queueserver.py
```diff
@@ -493,13 +477,11 @@ def _build_plan(self, suggestions: list[dict]) -> BPlan:
...
-    def _is_expected_start_doc(self, start_doc: RunStart) -> bool:
-        """Return True only for the acquisition currently expected by this runner."""
-        with self._state_lock:
-            return self._state is not None and self._state.current_uid == start_doc.get(CORRELATION_UID_KEY)
-
     def _on_acquisition_complete(self, start_doc: RunStart, stop_doc: RunStop) -> None:
         """Callback when acquisition finishes. Ingest results and maybe continue."""
+        with self._state_lock:
+            if self._state is None or self._state.current_uid != start_doc.get(CORRELATION_UID_KEY):
+                return
```
thopkins32: This needs to be moved back within `_process_acquisition` under the `try` block and `raise`. This way, the future gets properly updated with an exception.

The original was:

```python
 if self._state.current_uid != start_doc.get(CORRELATION_UID_KEY):
    raise RuntimeError(
        "current_uid did not match start document. "
        f"Got: {start_doc.get(CORRELATION_UID_KEY)}, Expected: {self._state.current_uid}"
    )
```
after the `RuntimeError("current_uid not set")` exception.

whs92: done

thopkins32 (approved):
~~Good to merge after one consideration.~~ I'll just merge as-is.

Feel free to merge with the docs build failing. I will address in a separate PR.

PR #318, opened 2026-06-04
Xopt integration

roussel-ryan:
This pull request adds support for integrating the Xopt optimization library into blop, introducing a new agent, optimizer, and mapping utilities, along with comprehensive tests for these components. The changes enable users to use Xopt Generator objects within the existing framework, and ensure robust mapping and interaction between the framework's abstractions and Xopt's API.

This PR should serve as a starting point for discussing Xopt integration into blop or bluesky-adaptive (https://github.com/bluesky/bluesky-adaptive/pull/61).

The most important changes are:

**Xopt Integration:**

* Added `XoptAgent`, `XoptOptimizer`, and `build_vocs` to the codebase, providing a new agent and optimizer that wrap Xopt generators and a utility for mapping framework abstractions to Xopt's VOCS format (`src/blop/xopt/agent.py`, `src/blop/xopt/optimizer.py`, `src/blop/xopt/mapping.py`, `src/blop/xopt/__init__.py`). [[1]](diffhunk://#diff-e568c7cc9069ce82c095259a18fbc4117e004bdfb9c2cab35cddd93629246b32R1-R187) [[2]](diffhunk://#diff-6565a0fec054de65018e8a3aa2441e2f49727f269e760f7986006728dcb06d55R1-R5)

* Made the new Xopt components available at the top-level API by importing them in `src/blop/__init__.py` and adding them to `__all__`. [[1]](diffhunk://#diff-86f76fdd9b7c6d484694dd3c9e0b4a10c2f8697973d5d16825a1e1aaf8bacb66R3) [[2]](diffhunk://#diff-86f76fdd9b7c6d484694dd3c9e0b4a10c2f8697973d5d16825a1e1aaf8bacb66R25-R27)

**Dependency Management:**

* Added `xopt` as an optional dependency in `pyproject.toml` to enable Xopt-based optimization.

**Testing and Validation:**

* Added comprehensive tests for the Xopt agent (`test_agent.py`), optimizer (`test_optimizer.py`), and VOCS mapping (`test_mapping.py`), covering initialization, suggestion, ingestion, checkpointing, constraint handling, and error cases. [[1]](diffhunk://#diff-a1fba18ed8403efb7339cb329e02fab14b77bb4b90701e79eb9a2c3c475fd294R1-R100) [[2]](diffhunk://#diff-dcdbb9d333ccdd466bc2c833339868d26f8240551f3501002ec5d20ba3cdc1dcR1-R224) [[3]](diffhunk://#diff-a7a1905370c05ba6efeb05a87ea026ca7165c4b81e238d45d0d299ca5e333a7eR1-R62)


src/blop/xopt/mapping.py
```diff
@@ -0,0 +1,168 @@
...
+) -> VOCS:
+    """Build an Xopt VOCS object from blop domain objects."""
+    # First build raw variable definitions from blop DOFs.
+    variables: dict[str, list[float] | list[float | int | str | bool]] = {}
+
+    for dof in dofs:
+        if isinstance(dof, RangeDOF):
+            variables[dof.parameter_name] = [float(dof.bounds[0]), float(dof.bounds[1])]
+        elif isinstance(dof, ChoiceDOF):
+            # Keep scope explicit: only numeric choices map cleanly to current VOCS variable model.
```
roussel-ryan: note that future development on the Xopt side will include discrete variables, see https://github.com/xopt-org/Xopt/pull/421

roussel-ryan:
Note: after discussion, I will refactor this PR to not do automatic conversion to VOCS parlance and just pass the generator object to an optimizer

roussel-ryan:
@thopkins32 so it looks like we need to make a decision since the more recent versions of botorch (.v0.17.0) require python >=3.11 meaning that Xopt does the same, how do you think we should resolve this? I could potentially re-enable support 3.10, but only if this is something you all think is necessary

thopkins32:
@roussel-ryan I will open a separate issue to remove Python 3.10 support, feel free to do that in this PR for now to unblock

whs92:
Can you give a quick overview of the advantages of using Xopt over AX? Why might I use one or the other? Or is there just an existing community using Xopt that we want to serve?

roussel-ryan:
Hi, there are several advantages to using Xopt over Ax for blop, primarily due to our focus on R&D towards accelerator / beam physics, however we do not intend to replace Ax completely. Some of the reasons:
- Xopt supports other optimization algorithms that can be used for online operation / optimization including Simplex, RCDS, Extremum Seeking, Genetic algorithms, and multiple variants of BO (most based on botorch, just as Ax is). We will shortly be adding methods for Reinforcement Learning and algorithms provided by scipy minimize
- We have added features to BO that are not included out of the box in Ax as it is currently implemented in blop, including first class support for trust region BO, time-dependent / contextual BO, multi-fidelity BO, interpolated measurements etc. (see https://journals.aps.org/prab/abstract/10.1103/PhysRevAccelBeams.27.084801)
- Xopt is wrapped by Badger https://github.com/xopt-org/Badger which provides a GUI interface for running optimization in control rooms. Utilizing Xopt generators in blop takes a step forward for creating a GUI that works for blop
- The same algorithms provided by Xopt used in online control via blop can be easily applied towards simulations and arbitrary problems, improving consistent use of algorithms across problems for users

Finally, supporting Xopt in blop more closely connects R&D efforts in accelerators and photon beamlines, allowing developments in one field to benefit the other. This PR is motivated by discussions this past week at the DSSI workshop for autonomous beamline control at BNL with @thopkins32 @tacaswell @mrakitin and others

whs92:
@roussel-ryan I am particularly interested in adding context to optimization. Thanks for your explanation.

roussel-ryan:
Good to hear, please keep an eye on https://github.com/xopt-org/Xopt/pull/425 which should be in the next release

whs92:
As far as I know adding additional observations for context was already supported by the earlier pre-AX version of blop. Whatever we come up with, blop should abstract away the differences (if they exist) between the different back ends so that if I want to provide context (or anything else) I shouldn't care whether I am talking to AX or XopT in the background.

roussel-ryan:
Based on previous discussions, we wanted to try to keep the Ax/Xopt interfaces separate (see previous comments above) and not include a translation layer to unify the interfaces to the user. I think we should discuss this further, but it may be a good initial step to get Xopt support capability integrated into blop first and then work on improving the user interface. I want to hear other folks opinions on the matter

roussel-ryan:
I've addressed all the issues on my side, I didn't make any changes to `tutorials/xrt-kb-mirrors.md` so I think that might be an issue that you all want to work on

src/blop/xopt/optimizer.py
```diff
@@ -0,0 +1,235 @@
...
+    ):
+        # Keep API simple: caller provides a fully configured Xopt generator instance.
+        self._generator = generator
+
+        # Internal state tracks IDs, pending/known parameterizations, and checkpoint metadata.
+        self._checkpoint_path = checkpoint_path
+        self._fixed_parameters: dict[str, Any] | None = None
+        self._next_id = 0
+        self._params_by_id: dict[int | str, dict[str, Any]] = {}
+        self._seed_state_from_existing_data()
```
thopkins32, line 43: Looking at the Xopt docs, is the data frame index of `self._generator.data` something we can use instead of doing ID tracking in Blop? I don't want to duplicate state tracking here if Xopt already has a mechanism for this.

The reason we have an ID associated with every suggestion is that when we evaluate multiple points in a single optimizer iteration (e.g., `RE(optimize(..., n_points=5))`) [we may have a more optimal sampling route than the default order provided by the optimizer](https://github.com/bluesky/blop/blob/90a01fae9557e10fa50dce0b49c19b20bd9788dc/src/blop/plans.py#L87-L92). Bluesky always emits events in the order they were collected ([and we cache this reordering information in the Bluesky metadata](https://github.com/bluesky/blop/blob/90a01fae9557e10fa50dce0b49c19b20bd9788dc/docs/source/tutorials/xrt-kb-mirrors.md?plain=1#L229-L230)). Therefore, to match Bluesky's event/collection order with the optimizer suggestion order, we need to track IDs for each suggestion.

thopkins32 (commented):
Initial review done. I'll follow up with a more in-depth one after I get clarification on one point.

thopkins32:
Also, I merged the change removing Python 3.10 support if you would like to rebase or merge `main` into this branch.

src/blop/xopt/optimizer.py
roussel-ryan: I would keep the ID tracking in Blop, certainly for now, but going forward as well. I think Xopt may adopt a more formal ID tracking formalism. Certainly we shouldn't hold up this PR to decide this IMO

src/blop/xopt/optimizer.py
```diff
@@ -0,0 +1,235 @@
...
+            if isinstance(trial_id, int):
+                self._next_id = max(self._next_id, trial_id + 1)
+
+    def suggest(self, num_points: int | None = None) -> list[dict]:
+        # Default to single-point suggestion when caller does not specify cardinality.
+        if num_points is None:
+            num_points = 1
+
+        # Delegate candidate generation to Xopt and optionally enforce fixed variables.
+        suggestions = self._generator.generate(num_points)
```
roussel-ryan: change to suggest / ingest nomenclature for xopt generator

src/blop/xopt/optimizer.py
```diff
@@ -0,0 +1,235 @@
...
+        if self._fixed_parameters:
```
roussel-ryan: need to look at integration with fixed parameters / VOCS / generator fixed values

roussel-ryan: may be able to remove

src/blop/xopt/optimizer.py
```diff
@@ -0,0 +1,235 @@
...
+            suggestions = [{**suggestion, **self._fixed_parameters} for suggestion in suggestions]
+        return self.register_suggestions(suggestions)
+
+    def register_suggestions(self, suggestions: list[dict]) -> list[dict]:
```
roussel-ryan, line 122: this data should be ingested into the generator

src/blop/xopt/optimizer.py
```diff
@@ -0,0 +1,235 @@
...
+        self._fixed_parameters: dict[str, Any] | None = None
+        self._next_id = 0
+        self._params_by_id: dict[int | str, dict[str, Any]] = {}
+        self._seed_state_from_existing_data()
+
+    @classmethod
+    def from_checkpoint(cls, checkpoint_path: str) -> "XoptOptimizer":
+        # Restore all persistent adapter state from pickle payload.
+        path = Path(checkpoint_path)
+        with path.open("rb") as stream:
```
roussel-ryan: look at proper serialization to json instead of using pickle

src/blop/xopt/optimizer.py
```diff
@@ -0,0 +1,235 @@
...
+            results.append((trial_id, parameterization, outcomes))
+
+        return results
+
+    def checkpoint(self) -> None:
+        # Enforce explicit checkpoint path configuration before writing state.
+        if not self._checkpoint_path:
+            raise ValueError("Checkpoint path is not set. Please set a checkpoint path when initializing the optimizer.")
+
+        # Persist generator and adapter bookkeeping to a single pickle artifact.
```
roussel-ryan: check proper serialization

src/blop/tests/xopt/test_optimizer.py
```diff
@@ -0,0 +1,382 @@
...
+
+    best_points = optimizer.get_best_points()
+    assert len(best_points) == 1
+    _, _, outcomes = best_points[0]
+    assert outcomes["y"] == 1.0
+    assert outcomes["obs"] == 20.0
+
+
+def test_xopt_expected_improvement_runs_simple_minimization():
+    vocs = VOCS(variables={"x": [0.0, 1.0]}, objectives={"y": "MINIMIZE"})
```
roussel-ryan, line 341: test inside run engine optimization plan

src/blop/xopt/optimizer.py
```diff
@@ -0,0 +1,235 @@
...
+            else:
+                trial_id = self._next_id
+                self._next_id += 1
+
+            self._params_by_id[trial_id] = {name: row[name] for name in self.vocs.variable_names if name in row}
+            if isinstance(trial_id, int):
+                self._next_id = max(self._next_id, trial_id + 1)
+
+    def suggest(self, num_points: int | None = None) -> list[dict]:
+        # Default to single-point suggestion when caller does not specify cardinality.
```
roussel-ryan, line 108: add in initial random sampling at first call / with no current data

roussel-ryan: add a potential option to specify type of random sampling at initializaiton

roussel-ryan (commented):
add tutorial (see https://github.com/bluesky/blop/blob/main/docs/source/tutorials/xrt-kb-mirrors.md)

src/blop/__init__.py
```diff
@@ -1,4 +1,5 @@
 from .plans import acquire_baseline, default_acquire, optimize, optimize_step, sample_suggestions
+from .xopt import XoptOptimizer
```
thopkins32: This assumes that someone installed `blop[xopt]`.
```suggestion
```

src/blop/__init__.py
```diff
@@ -12,4 +13,5 @@
     "optimize",
     "optimize_step",
     "sample_suggestions",
+    "XoptOptimizer",
```
thopkins32:
```suggestion
```

thopkins32 (changes requested):

thopkins32 (approved):

roussel-ryan:
@whs92 just as a followup the most recent version of Xopt (3.2.0) now supports contextual BO (see https://xopt.xopt.org/examples/single_objective_bayes_opt/contextual_bo/ for an example)

PR #321, opened 2026-06-08
Update & Rework Dependencies

### LINKED ISSUE #319: Bump dependencies versions
Integration tests are failing due to ophyd-async releases. Possibly from newer Tiled and ax-platform versions.

### LINKED ISSUE #320: Drop Python 3.10 support
It looks like BoTorch has dropped support for Python 3.10: https://github.com/meta-pytorch/botorch/blob/7a5692d5fe71e15595b88cb62c27dd821ea94c42/pyproject.toml#L12

We should probably do the same so we can stay up-to-date.

thopkins32:
With #318 , we will soon have two optimizer backends in Blop. Therefore, it is time to create a more modular dependency tree with optional extras for `ax` and `qs`.

The base dependencies of Blop now only refer to the essentials required for accessing the Bluesky plans and callbacks that nearly every Blop user would use.

If someone wants to use the Ax backend, they can install the `blop[ax]` extra.
If someone wants to use queueserver, they can install the `blop[qs]` extra.
Soon, they can choose to install `blop[xopt]` extra.

**Changing these dependencies to be extras creates some breaking changes to import paths.** Functionality-wise, nothing changes in this PR.

We also drop support for Python 3.10, since botorch does in their latest release.

Closes #320 
Closes #319

### DIFF
--- a/.github/workflows/_testing.yml
+++ b/.github/workflows/_testing.yml
@@ -10,7 +10,7 @@ jobs:
     strategy:
       matrix:
         host-os: ["ubuntu-latest"]
-        python-version: ["py310-cpu", "py311-cpu", "py312-cpu", "py313-cpu"]
+        python-version: ["py311-cpu", "py312-cpu", "py313-cpu"]
       fail-fast: false
 
     defaults:
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -21,6 +21,34 @@ To install the package using the ``conda`` package manager, run the following co
    :start-after: .. snippet-conda-standard-start
    :end-before: .. snippet-conda-standard-end
 
+Optional Extras
+^^^^^^^^^^^^^^^
+
+``blop`` is modular — install only what you need by appending one or more extras:
+
+.. list-table::
+   :header-rows: 1
+   :widths: auto
+
+   * - Extra
+     - Installs
+     - Notes
+   * - ``blop[ax]``
+     - ``ax-platform``, ``botorch``, ``gpytorch``, ``torch``
+     - GPU torch by default; pair with ``[cpu]`` for CPU-only
+   * - ``blop[queueserver]``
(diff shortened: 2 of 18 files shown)


MTakahashi-KWH (approved):
Functionally the same. This also addresses the last dangling "_id" to  ID_KEY concern.

would note it may be worth providing a blop[standalone], blop[aio], or blop[all] configuration, or perhaps a default packageless agent that randomly samples the configuration space. That would give users coming from Bluesky a single-import way to experiment with the library without immediately confronting them to choose and install a secondary package just to run Blop.

In general, i think any package configuration should support a standalone mode so its components can be explored and evaluated in isolation. (extremely helpful for new and less technical users, since most will first download blop or blop[all])

josephhanrahan:
I agree with @MTakahashi-KWH. Given this arrangement, having something like [base] and [all] becomes important for installation. Otherwise looks good.

josephhanrahan (approved):

thopkins32:
@josephhanrahan @MTakahashi-KWH Fair points. How about the following? This can go on the installation docs page.

| Extra | Installs | Notes |
|---|---|---|
| `blop[ax]` | `ax-platform`, `botorch`, `gpytorch`, `torch` | GPU torch by default; pair with `[cpu]` for CPU-only |
| `blop[xopt]` | `xopt` | Coming soon |
| `blop[scipy]` | `scipy` | Coming soon |
| `blop[queueserver]` | `bluesky-queueserver-api` | Transport layer only; pair with `[ax]` for `QueueserverAgent` |
| `blop[cpu]` | *(uv index routing)* | Routes `torch` to the CPU-only PyTorch index; requires `uv` |
| `blop[all]` | All backends + `[queueserver]` | No dev tooling; will grow as new backends are added |
| `blop[dev]` | `blop[all]` + dev tooling | For contributors |

PR #322, opened 2026-06-25
Alternate optimizers

### LINKED ISSUE #282: Add gradient based methods to default package capabilities
Currently the default optimizer is with the Ax agent, which uses a global optimization method and requires uniform sampling of the domain to prepare conditional and Gibbs sampling. This is powerful when there is not prior knowledge of the space (black box), but reduces relative performance for when prior assumptions can be made. 

This includes priors like:
- we already start in a "good enough" subspace (and bounds are fuzzy)
- we can make claims about the conditional distribution before starting (eg Gaussian centered around start point, or any other skewed/non isometric dist)
- Only a subset of parameters given actually affect the objective.

Ultimately, I suggest we add a gradient based optimizer to the default utilities so that we can target the use case where:
- ||solution-start|| <<  search space diameter
- the space is locally differentiable and a local minima is acceptable

 In other words, the optimal solution is a small perturbation of a high-dimensional starting point. This is good for problems associated with drift and optimizing a distribution quality rather than the distribution itself (ie reduce skew, make more gaussian, improve collimation ...)

MTakahashi-KWH:
#### Keeping in draft while the rest of the optimizers provided by scipy optimize are filled in on the SCP enum (should be pretty procedural with no changes needed to testing except regarding global optimizers)
_________
Request introduces a new Scipy-based gradient optimization interface to the Blop library and provides some documentation and usability improvements for gradient-based optimization workflows. The main changes include the addition of a user-friendly `Scipy` agent interface and standard independent `ScipyOptimizer` (with/session interface and standard object usage), new documentation for Scipy optimization, and enhancements to DOF support for Scipy optimizers.

**New Scipy optimization interface and documentation:**

* Rerigged `simple-experiment.md` as a system_spanning tutorial (`gradient-optimization.md`) demonstrating how to use Blop with Scipy optimizers, including setup, defining DOFs/objectives, evaluation functions, running optimizations, and visualizing results.
* Introduced a new `Scipy` class in `src/blop/gradient/Scipy.py` that provides a convenient, Ax-like agent interface (`Scipy.Agent(...)`) for running gradient-based optimizations with Scipy, supporting both single and multipoint sampling, callback management, and integration with Bluesky.
   * Domain native configuration interface (`ScipyCFG(...)` and `Scipy(ScipyCFG)`) also provided for deeper customization of scipy optimizers

**Enhancements to DOF/parameter handling:**

* Added a `to_scipy_bounds` method to the `RangeDOF` class, enabling easy conversion of DOFs to Scipy's `Bounds` object for compatibility with Scipy optimizers. (may be dropped due to 2 scipy optimizers unable to support bounds??)
* Imported Scipy's `Bounds` in `src/blop/ax/dof.py` to support the new conversion method.

### Targets:
#282 Add gradient based methods to default package capabilities

### DIFF
--- a/src/blop/ax/dof.py
+++ b/src/blop/ax/dof.py
@@ -126,9 +126,7 @@ class RangeDOF(DOF):
         )
 
     def to_scipy_bounds(self) -> Bounds:
-        """
-        convert DOF to the Scipy equivalent Bounds
-        """
+        """Convert DOF to the Scipy equivalent Bounds."""
         return Bounds(lb=self.bounds[0], ub=self.bounds[1])
 
 
--- a/src/blop/ax/dof.py
+++ b/src/blop/ax/dof.py
@@ -8,7 +8,6 @@ from typing import Literal, cast
 
 from ax import ChoiceParameterConfig, RangeParameterConfig
 from ax.api.types import TParameterValue
-from scipy.optimize import Bounds
 
 from ..protocols import Actuator
 
--- a/src/blop/ax/dof.py
+++ b/src/blop/ax/dof.py
@@ -8,6 +8,7 @@ from typing import Literal, cast
 
 from ax import ChoiceParameterConfig, RangeParameterConfig
 from ax.api.types import TParameterValue
+from scipy.optimize import Bounds
 
 from ..protocols import Actuator
 
@@ -123,6 +124,12 @@ class RangeDOF(DOF):
             step_size=self.step_size,
             scaling=self.scaling,
         )
+
+    def to_scipy_bounds(self) -> Bounds:
+        """
+        convert DOF to the Scipy equivalent Bounds
+        """
+        return Bounds(lb=self.bounds[0], ub=self.bounds[1])
 
 
 @dataclass(frozen=True, kw_only=True, eq=False)
--- a/src/blop/ax/dof.py
+++ b/src/blop/ax/dof.py
@@ -124,10 +124,6 @@ class RangeDOF(DOF):
             scaling=self.scaling,
         )
 
-    def to_scipy_bounds(self) -> Bounds:
-        """Convert DOF to the Scipy equivalent Bounds."""
-        return Bounds(lb=self.bounds[0], ub=self.bounds[1])
-
 
 @dataclass(frozen=True, kw_only=True, eq=False)
 class ChoiceDOF(DOF):
--- a/src/blop/gradient/Scipy.py
+++ b/src/blop/gradient/Scipy.py
@@ -1,0 +1,96 @@
+from collections.abc import Sequence
+from dataclasses import dataclass
+from threading import Event, thread
+from typing import Any, cast
+
+from scipy.optimize import Bounds, dual_annealing, minimize
+
+from blop.ax.dof import RangeDOF, DOFConstraint
+from blop.ax.objective import OutcomeConstraint
+
+from ..protocols import AcquisitionPlan, Actuator, EvaluationFunction, OptimizationProblem, Optimizer, Sensor
+
+
+@dataclass
+class ScpCFG:
+    dof: Sequence[RangeDOF]
+    # dof_constraints: Sequence[DOFConstraint] | None = None
+    outcome_constraints: Sequence[OutcomeConstraint] | None = None
+    Optimizer: str = "Default"
+    max_iter: int | None = None
+    eps: float | None = None
+
+
+class Scipy:
+    def __init__(
+        self,
+        sensors: Sequence[Sensor],
+        dofs: Sequence[RangeDOF],
+        evaluation_function: EvaluationFunction,
+        acquisition_plan: AcquisitionPlan | None = None,
+        # dof_constraints: Sequence[DOFConstraint] | None = None,
+        outcome_constraints: Sequence[OutcomeConstraint] | None = None,
+        checkpoint_path: str | None = None,
+        **kwargs: Any,
+    ):
+        self._sensors = sensors
+        self._actuators: Sequence[Actuator] = [cast(Actuator, dof.actuator) for dof in dofs if dof.actuator is not None]
+        self._evaluation_function = evaluation_function
+        self._acquisition_plan = acquisition_plan
+        self._params = []
+        self._bounds = 
+        for dof in dofs:
+            self._params.append(dof.parameter_name)
+            self.bounds.append(Bounds(lb=param.bounds[0], ub=param.bounds[1]))
+
+    @classmethod
+    def configure(cls, config: ScipyCFG):
+        self = cls()
+        self.cfg = config
+        return cls()
+
+    def to_optimization_problem(self) -> OptimizationProblem:
+        ...
+
+    def optimize():
+        ...
+
+
+class ScipyOptimizer(Optimizer):
+
+    def __init__(self, ScpCFG):
+        ...
+
+    def suggest(self, num_points: int | None = None) -> list[dict]:
+        """
+        Returns a set of points in the input space, to be evaulated next.
+
+        The "_id" key is optional and can be used to identify suggested trials for later evaluation
+        and ingestion.
+
+        Parameters
+        ----------
+        num_points : int | None, optional
+            The number of points to suggest. If not provided, will default to 1.
+
+        Returns
+        -------
+        list[dict]
+            A list of dictionaries, each containing a parameterization of a point to evaluate next.
+            Each dictionary must contain a unique "_id" key to identify each parameterization.
+        """
+        ...
+
+    def ingest(self, points: list[dict]) -> None:
+        """
+        Ingest a set of points into the experiment. Either from previously suggested points or from an external source.
+
+        The "_id" key is optional and can be used to identify points from previously suggested trials or to identify
+        the point as a "baseline" trial.
+
+        Parameters
+        ----------
+        points : list[dict]
+            A list of dictionaries, each containing the outcomes of each suggested parameterization.
+        """
+        ...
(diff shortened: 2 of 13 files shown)


MTakahashi-KWH:
This is now a mostly complete implementation. There are some unexposed optimizers and scipy options that I believe should be pushed to a future update of this module.

src/blop/scipy/scipy.py
```diff
@@ -0,0 +1,295 @@
+"""Scipy optimization power class for fast start QOL and Ax like agent behavior."""
```
thopkins32, line 1: Filename should be all lowercase.

src/blop/scipy/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""Scipy Backend for Pertubative gradient and in house global optimizers."""
```
thopkins32, line 1: sub-module should be named `scipy` not `gradient` to match the same pattern as `ax` and `xopt`.

src/blop/gradient/optimizer.py
```diff
@@ -0,0 +1,311 @@
...
+from dataclasses import dataclass
+from enum import StrEnum
+from threading import Thread
+from typing import Any, cast
+
+import numpy as np
+from scipy.optimize import OptimizeResult, dual_annealing, minimize, shgo
+
+from blop.ax.dof import RangeDOF
+from blop.ax.objective import Objective
```
thopkins32: Can't cross the backend boundary here now that we have modular dependencies. Someone may want to install scipy optimizers without installing Ax.

I suggest taking only what you need from RangeDOF and Objective and making custom scipy structures in this sub-module rather than importing the Ax backend ones. It isn't a lot of code.
```suggestion
```

src/blop/gradient/optimizer.py
```diff
@@ -0,0 +1,311 @@
...
+    COBYQA = "COBYQA"
+    SLSQP = "SLSQP"
+    Trust_Constr = "trust-constr"
+    # Dogleg = "dogleg"
+    # Trust_NCG = "trust-ncg"
+    # Trust_Exact = "trust-exact"
+    # Trust_Krylov = "trust-krylov"
+
+    Dual_Annealing = "dual annealing"
+    SHGO = "SHGO"
```
thopkins32: Enums must be all caps, surprised the ruff linter didn't catch this.

src/blop/gradient/Scipy.py
```diff
@@ -0,0 +1,295 @@
+"""Scipy optimization power class for fast start QOL and Ax like agent behavior."""
+
+from collections.abc import Mapping, Sequence
+from typing import Any, cast
+
+import bluesky.preprocessors as bpp
+from bluesky.callbacks import CallbackBase
+
+from blop.ax.dof import RangeDOF
+from blop.ax.objective import Objective
```
thopkins32: Same thing here as previous comment, can't import Ax in scipy backend code.

src/blop/ax/dof.py
```diff
@@ -124,6 +125,10 @@ def to_ax_parameter_config(self) -> RangeParameterConfig:
             scaling=self.scaling,
         )
 
+    def to_scipy_bounds(self) -> Bounds:
+        """Convert DOF to the Scipy equivalent Bounds."""
+        return Bounds(lb=self.bounds[0], ub=self.bounds[1])
+
```
thopkins32:
```suggestion
```

src/blop/ax/dof.py
```diff
@@ -8,6 +8,7 @@
 
 from ax import ChoiceParameterConfig, RangeParameterConfig
 from ax.api.types import TParameterValue
+from scipy.optimize import Bounds
```
thopkins32:
```suggestion
```

src/blop/tests/gradient/test_integration.py
```diff
@@ -0,0 +1,42 @@
...
+    agent = Scipy(
+        sensors=[readable],
+        config=config,
+        evaluation_function=deflating_evaluation(),
+        timeout=5,
+    )
+    agent._optimizer.force_resiliance = True
+    RE = RunEngine({})
+    RE(agent.optimize(20))
+    time.sleep(0.1)
```
thopkins32: Tests must pass without arbitrary sleeps
```suggestion
```

src/blop/scipy/optimizer.py
```diff
@@ -0,0 +1,311 @@
...
+        Parameters
+        ----------
+        points : list[dict]
+            A list of dictionaries, each containing the outcomes of each suggested parameterization.
+        """
+        for res in points:
+            y = res[self._objective.name]
+            if res[ID_KEY] not in self._active:
+                if not self.force_resiliance:
+                    raise ValueError("optimizer did not expect to receive an update")
```
thopkins32, line 222: I am confused by this. Why is it necessary?

src/blop/scipy/optimizer.py
```diff
@@ -0,0 +1,311 @@
...
+                    func=cost,
+                    x0=_x,
+                    bounds=self._bounds,
+                    callback=dual_callback,
+                    minimizer_kwargs={"callback": default_callback, "bounds": self._bounds, "options": kws},
+                )
+        elif config.optimizer is SCP.SHGO:
+            # TODO the utility of SHGO is quite underepresented in this implementation, much more thought needs to go into
+            # how parameters are passed through this formalism
+            print("warning: as a global optimizer, SHGO does not use an X0 but its own Sobol sampling")
```
thopkins32, line 114: Use `logger.warning` instead of bare `print`.

src/blop/scipy/optimizer.py
```diff
@@ -0,0 +1,311 @@
...
+
+            def shgo_callback(x):
+                print(f"callback point {x} with current best of {self.intermediate}")
+                # self.intermediate = self.Result(x, -1, self._increment, 1)
```
thopkins32, line 118: Is this still necessary?

thopkins32 (changes requested):
Fairly complex implementation so I will have to review once more after discussion. Are there practical ways to simplify a lot of this?

The major complexity stems from the required cooperative thread + callback mechanism and the support for various algorithms from `scipy.optimize`. That + some state tracking should be all we need to get an optimizer working with Blop's protocols.

If we defer to the user to give us the `scipy.optimize` method (w/ `*args`, `**kwargs`) in the scipy Agent/Optimizer, then all we have to do is state tracking + cooperative thread management.

Our "cost function", in scipy terms, is *always* to simply wait (possibly forever) until the next call to `ingest()`, nothing more, nothing less.

src/blop/gradient/optimizer.py
thopkins32: See my comment on API design for this. Can users pass in the `scipy.optimize` method they want instead of us maintaining a "supported algorithm" list?

src/blop/gradient/optimizer.py
MTakahashi-KWH: Am happy to delete this. This was some dangling code from when i first implemented this and thought that optimization problem components may be a more concrete portable API

src/blop/gradient/optimizer.py
MTakahashi-KWH: This is a debate I've have with development as Scipy is notorious for having an inconsistent API. I decided to 'flatten' the call so that a standard parameter/call format fits them all, and needed enum as a gatekeeper for as I expand the implementation support. Adding the ability to pass an optimizer would would need a standard enough parameter interface Scipy.opt doesnt have (see some algorithms also requiring a jacobian or an analytical gradient closure too)

In a sense my implementations primary development is to invert the prompting loop of a cost closure calling optimization function and not necessarily scipy specific. I can definitely make a generic loop inversion implementation and regularize scipy's calls in another file to work with this and we'd just pass one of those as something like "loop_invert(scipy_norm.BFGS,*params)". Another optimization system that accepts a cost function closure could run the same way. Though ultimately that's a consequence of the implementation, not the objective. The design started from the assumption that SciPy is the optimization backend to enable gradient optimization, not from a desire to abstract over arbitrary optimizers. I see no significant evidence that another optimization library operates in a similar way to enable this.

src/blop/tests/gradient/test_integration.py
MTakahashi-KWH: I double check why this is here. It may be from a race condition regarding optimizer.final and optimizer_active

src/blop/scipy/optimizer.py
MTakahashi-KWH: This is from the fact that we don't have a stopping condition and agent.optimize() runs for a fixed number of iters. If scipy converges to a solution, it will stop creating sample points. I made the decision that if it converges in the middle of an optimization, a user would be more happy for it to run the iterations to completion without erroring out run engine. But I still give the ability to do so anyways to cut useless operation.

src/blop/gradient/optimizer.py
thopkins32: Yeah I realize this is a bit annoying. Most of these methods take a `callable` function as the first argument though which is something we can standardize on. In fact, all of the methods under the "Optimization" sub-header take `func` as the first argument.

This would enable the use of:
- `minimize_scalar`
- `minimize` with any choice of method
- `basinhopping`
- `brute`
- `differential_evolution`
- `shgo`
- `dual_annealing`
- `direct`

This already covers so much that I think it's fine to start here. If there is demand for anything beyond this, then we can revisit.

MTakahashi-KWH:
added refactoring branch [scipy_barebones](https://github.com/bluesky/blop/tree/scipy_barebones) which separates the parameter normalization of inner/closed optimizer call (InnerOptimizer) from the inversion setup (OuterOptimizer). Really need a better naming convention for this structure

of interest to perhaps:
@thopkins32 
@josephhanrahan

PR #326, opened 2026-07-18
Use proper autostart API for queueserver

### LINKED ISSUE #314: Queueserver autostart should use the proper API
Right now we wait for idle or paused and then start the queue. But really we should simply use https://blueskyproject.io/bluesky-queueserver-api/generated/bluesky_queueserver_api.zmq.REManagerAPI.queue_autostart.html#bluesky_queueserver_api.zmq.REManagerAPI.queue_autostart

thopkins32:
Closes #314 

Discovered while testing at XPD. This PR is a simple fix that uses the proper autostart API for queueserver rather than inventing a custom one on each plan submission.


src/blop/queueserver.py
```diff
@@ -188,28 +193,18 @@ def check_plan_available(self, plan_name: str) -> None:
...
-            If True, start the queue after adding the plan.
-        timeout : float, optional
-            Timeout in seconds when waiting for queue to be idle.
         """
         response = self._rm.item_add(plan)
         logger.debug(f"Submitted plan to queue. Response: {response}")
 
-        if autostart:
-            logger.debug("Waiting for queue to be idle or paused")
-            self._rm.wait_for_idle_or_paused(timeout=timeout)
```
MTakahashi-KWH, line 209: could you give a short explanation for why the queue needs to be restarted? id assume once the queue item has been added, that the queue will become idle only after processing the submitted item, why wait? Moreover wouldn't it be more likely a wait for idle or pause be done before task submissions?

thopkins32: > id assume once the queue item has been added, that the queue will become idle only after processing the submitted item, why wait?

This is an incorrect assumption. If the proper autostart API is used and set to true, then this assumption holds. If not, then the queue will not start until explicitly told to do so. Otherwise, you may be running someone else's plan that was before you in the queue when you add your item.

Anyway, this code was removed because it was incorrect. This PR aims to address that.

Reference: https://blueskyproject.io/bluesky-queueserver-api/generated/bluesky_queueserver_api.zmq.REManagerAPI.queue_autostart.html#bluesky_queueserver_api.zmq.REManagerAPI.queue_autostart

MTakahashi-KWH: What a weird default behavior. The new autostart is a lot better imo.

MTakahashi-KWH (approved):
approved, removed awkward state control of queue by users through use of autostart

PR #334, opened 2026-07-31
Scipy barebones

MTakahashi-KWH:
Revisions to  the Scipy pull request to better separate the optimizer wrappers from the cooperative inversion. likely more work to do regarding demoing how to use optimizer directly. Weird side note is the optimizer object can run in a classic mode now by only using the normalized interface (also might be cool to explore with inversion of badger?)


MTakahashi-KWH:
should i add the brute, minimize_scale, univariate , basin_hopping, and diff_evo to this before merge?

src/blop/tests/scipy/test_integration.py
```diff
@@ -29,12 +31,12 @@ def __call__(self, uid, suggestions):
...
-    # time.sleep(0.1)
-    assert agent._optimizer.intermediate is not None
-    assert not agent._optimizer._active
-    RE(agent.optimize(20))
-    assert agent._optimizer.final is not None
+    assert agent.optimizer.intermediate is not None
+    assert not agent.optimizer._active
+    RE(agent.optimize(40))
     assert agent.get_best_points() is not None
+    time.sleep(0.1)
```
thopkins32: Must remove this sleep. Why is it needed?
```suggestion
```

thopkins32 (changes requested):
Still need to remove the sleep in tests. If it's still needed, it indicates a race condition somewhere.

src/blop/tests/scipy/test_integration.py
MTakahashi-KWH: lol its a race condition. commit #3869467 touches this its due to a slow closing of the internal optimizer when returning final result. Final is not a guaranteed API property of the object and its data access essentially would be abstracted through get_best_points. Ill just drop it as an assertion

PR #337, opened 2026-08-03
add GlobalStoppingAware protocol

jessica-moylan:
The start of adding a global early stopping condition.

### DIFF
--- a/src/blop/plans.py
+++ b/src/blop/plans.py
@@ -15,9 +15,9 @@ from .protocols import (
     ID_KEY,
     Actuator,
     CanRegisterSuggestions,
-    GlobalStoppingAware,
     OptimizationProblem,
     Sensor,
+    StoppingConditions,
     TrialFaultAware,
 )
 from .utils import InferredReadable, _maybe_checkpoint, collect_optimization_metadata, route_suggestions
@@ -168,7 +168,6 @@ def optimize(
     n_points: int = 1,
     checkpoint_interval: int | None = None,
     readable_cache: dict[str, InferredReadable] | None = None,
-    global_stopper: GlobalStoppingAware | None = None,
     **kwargs: Any,
 ) -> MsgGenerator[None]:
     """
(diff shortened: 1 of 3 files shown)


src/blop/plans.py
```diff
@@ -167,6 +168,7 @@ def optimize(
     n_points: int = 1,
     checkpoint_interval: int | None = None,
     readable_cache: dict[str, InferredReadable] | None = None,
+    global_stopper: GlobalStoppingAware | None = None,
```
thopkins32: This is already in the optimization problem
```suggestion
```

src/blop/plans.py
```diff
@@ -187,6 +189,8 @@ def optimize(
     readable_cache: dict[str, InferredReadable] | None = None
         Cache of readable objects to store the suggestions and outcomes as events.
         If None, a new cache will be created.
+    global_stopper: GlobalStoppingAware | None = None,
+        The global stopping strategy to determine when the optimization should stop.
```
thopkins32:
```suggestion
```

src/blop/plans.py
```diff
@@ -218,6 +222,12 @@ def _optimize() -> MsgGenerator[None]:
             # Perform a single step of the optimization
             uid, suggestions, outcomes = yield from optimize_step(optimization_problem, n_points, **kwargs)
 
+            if isinstance(optimization_problem.optimizer, GlobalStoppingAware):
+                stop_now, stop_reason = optimization_problem.optimizer.should_stop()
+                if stop_now:
+                    print(f"Global stopping triggered at iteration {i + 1}: {stop_reason}")
```
thopkins32:
```suggestion
                    logger.info(f"Global stopping triggered at iteration {i + 1}: {stop_reason}")
```

thopkins32: What if `stop_reason` is `None`?

thopkins32 (changes requested):
A few suggestions.

I would also suggest removing "global" from the protocol. Ax may have "global" stopping criteria, but that wording may not apply to all optimizer backends.

Please also add tests with a mocked protocol interface.

MTakahashi-KWH:
Would using "...aware" be a misnomer as the optimizer would be emitting the global stop command/signal. I would think something like "GlobalStopSignaling/able" if using mixin naming or "GlobalStopSignal" if using protocol object naming would be clearer.

thopkins32 (approved):
LGTM, may suggest a follow-up name change to `HasStoppingCriteria` or `SupportsStoppingCondition` but functionally this looks correct.

PR #338, opened 2026-08-03
Rename StoppingConditions to SupportsStoppingCriteria

thopkins32:
Renaming `StoppingConditions` to be closer to the style of other protocol names like `CanRegisterSuggestions`.


jessica-moylan (approved):
LGTM. I agree this renaming sounds cleaner

PR #342, opened 2026-08-13
Stopping Criteria documentation

jessica-moylan:
Added documentation for how to use `SupportsStoppingCriteria` as well as how to add custom methods

Added two stopping conditions with tests

- `MaxEvaluationsStopping` which stops after a set number of iterations
- `ValidConfigurationStopping` which stops once it finds a feasible configuration after a minimum number of iterations


src/blop/ax/stopping_conditions.py
```diff
@@ -0,0 +1,65 @@
+"""Stopping conditions that can be used to halt the optimization process when using the Ax backend."""
+
+from ax.core.experiment import Experiment
+from ax.global_stopping.strategies import BaseGlobalStoppingStrategy
+from ax.global_stopping.strategies.improvement import constraint_satisfaction
```
thopkins32: We don't need custom stopping strategies in Blop. Let's leave it to Ax to offer them. ~~They have good coverage on many of them here: https://github.com/facebook/Ax/blob/6e711c86a162f8fc5173b0d5e18f0326c329bac7/ax/early_stopping/strategies/__init__.py#L9-L22~~

I linked early stopping, sorry.

https://github.com/facebook/Ax/blob/6e711c86a162f8fc5173b0d5e18f0326c329bac7/ax/global_stopping/strategies/__init__.py#L13-L16 are the global stopping strategies offered by Ax.

thopkins32 (changes requested):
Stopping strategy implementations are already offered by Ax. Feel free to add those to the documentation in this PR.

https://github.com/facebook/Ax/blob/6e711c86a162f8fc5173b0d5e18f0326c329bac7/ax/global_stopping/strategies/__init__.py#L13-L16

src/blop/ax/stopping_conditions.py
MTakahashi-KWH: The documentation looks good to me but agreed on custom stopping strategies. 

I think since we already inherit this protocol as a tack on interface and with the breadth of stopping reasons largely means that any implementation of the protocol intends to make a unique stopping descriptor (ie convergence, too much landscape variance, iteration). I don't think its obligate or necessary for blop core to know/include structure and reason behind the stop.

thopkins32: Agreed, we should try to contribute them upstream in Ax rather than keep them here, if we come up with new ones that aren't already covered.

docs/source/how-to-guides/stopping-conditions.rst
```diff
@@ -0,0 +1,108 @@
...
+        def __init__(self, name: str, initial_value: float = 0.0) -> None:
+            super().__init__(name)
+            self._value: float = initial_value
+
+        def set(self, value: float) -> Status:
+            self._value = value
+            return AlwaysSuccessfulStatus()
+
+    db = MagicMock(spec=Container)
+
```
thopkins32:
```suggestion
```

thopkins32 (approved):

PR #349, opened 2026-08-21
Add integration test suite and github workflow

### LINKED ISSUE #348: Integration tests on a schedule
Right now, we only do proper integration testing using the docs build. We should have a proper pytest suite that does integration testing that runs on a schedule (once-per-day or once-per-week).

This can help us validate the assumptions we make about our dependencies in our existing unit test suite. The reason I would like to keep the separate is to keep unit tests very very fast. Integration tests can be much slower.

thopkins32:
Assisted by: oh-my-pi:gpt-5.5

# Summary

Adds integration testing suite.

Closes #348 

To be expanded upon in future work, as needed.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -8,6 +8,9 @@ omit =
     # ignore _version.py and versioneer.py
     .*version.*
     *_version.py
+    # ignore test modules
+    src/blop/tests/**
+    */blop/tests/**
 
 exclude_lines =
     if __name__ == '__main__':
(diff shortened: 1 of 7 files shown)


MTakahashi-KWH (approved):
Smoke test that catches bad changes in internal interfaces. Detached from regular push/ pull request triggering and set to trigger on 6am cron "schedule" event. Looks good

Issue #336, opened 2026-08-01
Support single-run streaming/callback optimization mode


thopkins32:
# Feature Request

Blop currently assumes each optimization step produces a separate acquisition run UID, which is then passed to the evaluation function. This works well for wrapping existing beamline acquisition plans, but it makes it difficult to represent fast or tightly integrated optimization workflows as a single Bluesky run.

Add an optional execution mode where optimization trials can be evaluated from references to data within an open Bluesky run, rather than from child acquisition UIDs.

# Proposed Direction

This could be a new evaluation API (protocol) that can work with either:
- Tiled streaming references to event/stream slices
- Raw Bluesky document caches collected during execution

The reason that it can't work with the existing `blop.protocols.EvaluationFunction` is because its contract requires a Bluesky run UID, which is only produced upon `'open_run'` and `'close_run'`.

In the implementation of an `optimize`-like plan for this, we would have to check that the user supplied acquisition plan does not contain `'open_run'` or `'close_run'` messages. It must be a "plan stub".
# Use Cases

- Single-run optimization with separate streams for suggestions, detector data, outcomes
- Fast control optimization where opening one run per iteration is unnecessary overhead

MTakahashi-KWH:
i think that blop's implementation is still open enough that it only expects a "data reference"(uid mostly though direct file path could be implicit too) rather then specifically a run_uid. We should be good as each event document has its own uid and we could fetch it from tiled or alternate cache. 

I tried making a non tiled cache system a while ago (to remove tiled as an implicit dependency) and posted it in #300 with the direct_link implementation as a similar implementation to the logger. 

I think if we expound that the aquisition plan just/must return a uid data reference which uniquely identifies stored data (either a full run or an event document uid ... not important to blop as its data irresponsible) its once again the responsibility of the implementer/evaluation to collect it from the data store and know its result structuring.

a minor extension of the implementation from that issue would be like
```
class OptimizationCache(CallbackBase):
    def __init__(self):
        self.run_log = {}
        self.sample_seq = {}

    def start(self, doc: RunStart):
        bs = doc.get('blop_suggestions', None)
        if bs is None:
            print("wrong start doc")
            return
        self.run_uid = doc.get("uid", None)
        print("run uid: ", self.run_uid)
        self.ids = [s[ID_KEY] for s in bs]

    def event(self, doc: Event) -> Event:
        sn = doc.get("seq_num", None)
        if sn is None:
            raise KeyError
[...]

PR #350, opened 2026-08-25
Protocol type relaxation

thopkins32:
# Summary

Relaxes the typing of core `blop.protocols` to accept Python `collections.abc` types.

# Motivation

Protocols should use general types in their method signatures rather than concrete Python types. This allows for greater flexibility in the protocol and less type conversion necessary for the user to deal with when implementing custom optimizers, evaluation functions, and acquisition plans.

One additional benefit is that these more generic types are immutable. E.g. `list` is mutable, but `Sequence` is not. Same with `dict` versus `Mapping`. Immutability is important to preserve and now the type checker can help us from making mistakes like this.

# Key Decisions

- `list[dict]` becomes `Sequence[Mapping]` (no change in default behavior)
- `uid: str` in `EvaluationFunction` becomes `uid: Hashable` (no change in default behavior, but added flexibility things such as `tuple[str, ...]` or custom hashable objects.
  - This change had downstream impacts on how we store this data in event-model through `InferredReadable`. I try to store the native type, unless it is a custom hashable object, where we store `repr(uid)` instead.
  - Makes development of #336 much easier since we can reference a tuple of event UIDs or something similar as the default acquisition behavior.
  - Generally gives greater flexibility for writing custom `AcquisitionPlan`s and `EvaluationFunction`s and fetching acquired data from storage.

# Breaking Changes

Users of Blop should not see any immediate breaking changes since we are merely relaxing type constraints. However, they may notice two things:
- Type checkers like `pyright` will fail if users write custom `AcquisitionPlan`s or `EvaluationFunction`s using the old types in the signature. This does not impact runtime behavior.
- Change from `"bluesky_uid"` as the data key for `optimization` run/stream event documents to `"acquisition_uid"` since the return type of `AcquisitionPlan` is now any `Hashable` object. It is no longer always a Bluesky run UID.

### DIFF
--- a/src/blop/tests/test_plans.py
+++ b/src/blop/tests/test_plans.py
@@ -99,12 +99,12 @@ def test_optimize(RE):
     assert len(events) == 1
     data = events[0]["data"]
     assert "suggestion_ids" in data
-    assert "bluesky_uid" in data
+    assert "acquisition_uid" in data
     assert "x1" in data
     assert "objective" in data
     assert data["x1"] == 0.0
     assert data["objective"] == 0.0
-    assert data["bluesky_uid"] and isinstance(data["bluesky_uid"], str)
+    assert data["acquisition_uid"] and isinstance(data["acquisition_uid"], str)
 
 
 def test_optimization_failure(RE):
@@ -165,7 +165,7 @@ def test_optimize_multiple(RE):
     for event in events:
         data = event["data"]
         assert "suggestion_ids" in data
-        assert "bluesky_uid" in data
+        assert "acquisition_uid" in data
         assert "x1" in data
         assert "objective" in data
         assert data["x1"] == 0.0
@@ -201,7 +201,7 @@ def test_optimize_multiple_with_n_points(RE):
     for event in events:
         data = event["data"]
         assert "suggestion_ids" in data
-        assert "bluesky_uid" in data
+        assert "acquisition_uid" in data
         assert "x1" in data
         assert "objective" in data
         sid = data["suggestion_ids"]
@@ -268,7 +268,7 @@ def test_optimize_complex_case(RE):
     for event in events:
         data = event["data"]
         assert "suggestion_ids" in data
-        assert "bluesky_uid" in data
+        assert "acquisition_uid" in data
         assert "x1" in data
         assert "x2" in data
         assert "x3" in data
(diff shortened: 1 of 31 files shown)


src/blop/tests/test_plans.py
```diff
@@ -74,12 +99,12 @@ def test_optimize(RE):
...
     data = events[0]["data"]
     assert "suggestion_ids" in data
-    assert "bluesky_uid" in data
+    assert "acquisition_uid" in data
     assert "x1" in data
     assert "objective" in data
     assert data["x1"] == 0.0
     assert data["objective"] == 0.0
-    assert data["bluesky_uid"] and isinstance(data["bluesky_uid"], str)
+    assert data["acquisition_uid"] and isinstance(data["acquisition_uid"], str)
```
MTakahashi-KWH, line 107: perhaps shift isinstance assertion to be the new hashable type so that changes to default acquires return type can vary with further implementation

i think in the event documents we already for convenience uplift an array type out of default uid space for inferring sample sectioned returns 

in read_step normalize_uid comes from:
```
 def _acquisition_identifier_value(uid: Hashable) -> ArrayLike:
     """Convert a hashable acquisition identifier to an event-readable value."""
     if _is_array_like_identifier(uid):
         return cast(ArrayLike, uid)
     return repr(uid)
```

thopkins32: `default_acquire` must continue to return only the Bluesky run UID, otherwise that's another, more serious, breaking change to the API. I don't really see a case where you would want to return something else for `default_acquire` that wouldn't be better served by writing your own custom acquisition function. Future refactoring can make the contents of `default_acquire` more re-usable, if that's what you're after.

The assertion is for `default_acquire`, specifically, so the `isinstance` check for `str` is intended here.

MTakahashi-KWH: Fair, that was an oversight by me. Though, i'd then say its not particularly clear or clean to enforce behavior of the default acquire through a test of optimize.

MTakahashi-KWH (approved):
all unit tests passing, with functional behavior preserved. Perhaps future edits to unit tests to better separate concerns, if auditability of unit tests is somehow a future priority (idk theres probably some documentation philosophy around displaying unit tests for what a code should do rather how a code does it with bare API reference)

josephhanrahan (approved):

Issue #300, opened 2026-05-13
QOL code to building evaluation functions for more simply framed optimizations?


MTakahashi-KWH:
## Premise
A subset of Bluesky RunEngine optimization streams should be well-structured within the event model (i.e. data lives directly in event documents) and acquisition plans (default_acquire). Because of this, we can provide reusable default implementations for evaluation-function data handling using common stream/data-store patterns. The goal is to providing utility to avoid "reinventing" the wheel every time they want to do a simple opt without taking on strict internal API guarantees around data handling.

I've been experimenting with decorators to separate data handling from explicit per sample data processing, also enabling code separation between:
* run-to-run configuration changes
* system/setup-level configuration

For simple or tightly integrated systems, this could make fast-start optimization workflows much cleaner.

## Example

```
dofs = [
    RangeDOF(actuator=motor1, bounds=(-3, 3), parameter_type="float"),
    RangeDOF(actuator=motor2, bounds=(-3, 3), parameter_type="float"),
    RangeDOF(actuator=motor3, bounds=(-3, 3), parameter_type="float"),
    RangeDOF(actuator=motor4, bounds=(-3, 3), parameter_type="float"),
]
objectives = [
    Objective(name="beam_peak", minimize=False),
]
readables = [det]

//////////////////////////// new:

@tiled_link(tiled_client,channels=[det])
def cost(measures):
    evaluation = 5*np.exp(-.1*measures[det.name])+.1
    return {"beam_peak" : evaluation}

or 

\\ exists as a proof of concept to decouple Tiled as core data store and feasibly use another within a conv func
stream_store = OptimizationCache()
RE.subscribe(stream_store)

@direct_link(stream_store=stream_store)
def cost(measures):
    evaluation = 5*np.exp(-.1*measures[det.name])+.1
    return {"beam_peak" : evaluation}

//////////////////////////////


agent = Agent(
    sensors=readables,
    dofs=dofs,
    objectives=objectives,
    evaluation_function=cost.linked,
    name="simple-experiment",
    description="A simple experiment optimizing the beam peak intensity",
)
RE(agent.optimize(iterations=4,n_points=3))
```

## Note:
The `Eval-function-wrapper` branch contains the current prototype. I'll be adding examples and comments over the next few days.

[...]

Issue #229, opened 2026-01-12
Easier use of document stream in evaluation


thopkins32:
Currently, the `EvaluationFunction` protocol makes some assumptions about how you retrieve the acquired data that was just collected. We should explore what it may look like to access data in-process rather than relying on Tiled or some other service to fetch the data.

You could technically do this now by ignoring the `uid` argument and writing something like:
```python
cache = []
def document_cache_callback(...):
    """Cache the documents of each run"""
    ...

RE.subscribe(document_cache_callback)

def custom_evaluation(uid: str, suggestions: list[dict]) -> list[dict]:
    """Use the cache instead of uid"""
    ...
```

It's a bit awkward to do so maybe there is something Blop can do to make this simple.

PR #353, opened 2026-08-25
Optimization within a single Bluesky run

### LINKED ISSUE #336: Support single-run streaming/callback optimization mode

thopkins32:
# Summary

Minimal implementation of a plan that does the optimization loop within a single run, without altering Blop's core protocols or inventing new ones. This is unlike `blop.plans.optimize` which opens Bluesky runs in parallel (one outer run, many inner runs). Now, there is only one single run, and the optimization data is collected into a separate event stream called "optimization".

This PR mostly sets the stage for future work that will make single-run optimization easier to use.

Closes #336 

# Detailed Changes

- New plan: `blop.plans.optimize_in_run` which implements the optimization loop (similar to `blop.plans.optimize` and `blop.plans.optimize_step`).
- New stub: `blop.plan_stubs.list_scan_in_run` which is a version of list scan (similar to `blop.plans.default_acquire`) that strips out `'open_run'` and `'close_run'` messages normally created by the `bluesky.plans.list_scan`.
  - I intentionally avoided using `bluesky.preprocessors.stub_wrapper` here because that also strips `'stage'` and `'unstage'` messages. Potentially lengthy operations such as the evaluation function, outcome ingestion, or next suggestions may require that devices be unstaged.
- Validate that suggestion IDs are unique and `Hashable` both after `optimizer.suggest()` and evaluation.
- Moved around utilities and helper plan stubs to better file locations
- Various updates to the docs explaining the relationship between acquisition plans and evaluation functions, grounded in this new example of optimization within a single run.

# Future Work

This PR opens the door to a significant chunk of potential future work. Once this PR is merged, we should open these as Issues.
- Refactoring common operations between `blop.plans.optimize`, `blop.plans.optimize_in_run`, `blop.plans.optimize_step`, `blop.plans.default_acquire`, and `blop.plan_stubs.list_scan_in_run`. As is, this PR will duplicate some code in some places.
- Blop's callback router should support `optimize_in_run` by allowing the `optimization` stream to pass through the filter.
  - A new logging callback may be nice for `optimize_in_run` if the current one doesn't fit or refactor nicely.
- A version of `blop.plans.optimize_in_run` that stages/unstages devices only once for the entire run. Current design is to stage/unstage every iteration. This behavior should be opt-in and explicitly documented.
- Expose `blop.plans.optimize_in_run` in the Ax agent (not the queueserver one)
- In the spirit of #229 and #300 , write evaluation function helpers or base classes that would assist users with juggling IDs between acquisition plans and evaluation functions and fetching their data
  - Document callback one that keeps an in-memory store of documents
  - Tiled streaming one that sets up subscriptions using the Tiled client
  - These likely can't apply to every possible situation, but we should try to target the most common cases, such as:
     - One suggestion <-> one Bluesky event
     - One suggestion <-> one Tiled client update (from Tiled subscriptions)

### DIFF
--- a/docs/source/explanations/acquisition-plan.rst
+++ b/docs/source/explanations/acquisition-plan.rst
@@ -26,8 +26,10 @@ chosen storage or event system (typically stashing the suggestions in proper
 order in storage or tagging the run markdown).
 
 The plan must return a hashable acquisition identifier. Blop passes that value
-unchanged to the evaluation function. A Bluesky run UID is the usual identifier,
-but a tuple of event UIDs can identify acquisitions performed within one run.
+unchanged to the evaluation function. A Bluesky run UID is the usual identifier
+for run-owning acquisitions. The in-run default returns suggestion IDs in the
+order they were executed, but custom plans may return any hashable key their
+matching evaluation function understands.
 
 A simple example that optimizes in a subspace while executing measurements in
 the full physical coordinate system is shown below. 
--- a/docs/source/explanations/evaluation-function.rst
+++ b/docs/source/explanations/evaluation-function.rst
@@ -12,7 +12,7 @@ commonly used at beamlines while also supporting data acquired within a single B
 Anatomy of an Evaluation Function
 ---------------------------------
 
-An evaluation function is a callable that accepts a hashable acquisition identifier and a sequence of suggestion mappings, then returns a sequence of outcome mappings. The identifier may be a Bluesky run UID, a tuple of event UIDs, or another hashable key understood by the evaluator.
+An evaluation function is a callable that accepts a hashable acquisition identifier and a sequence of suggestion mappings, then returns a sequence of outcome mappings. The identifier may be a Bluesky run UID, suggestion IDs in executed order, a tuple of event UIDs, or another hashable key understood by the evaluator. The suggestion sequence is optimizer-provided and is not guaranteed to be in acquisition order; match data and outcomes by ``_id``.
 
 A typical implementation is shown below:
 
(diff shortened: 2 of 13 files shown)


thopkins32:
@josephhanrahan @MTakahashi-KWH ~~Looking through my PR again from a fresh perspective, do you think we need a separate `optimize_in_run` plan or can we simply use `optimize` with `list_scan_in_run`?~~

Never mind, I think keeping them separate makes sense. But we should refactor the common bits in future work.

MTakahashi-KWH:
have you tried this with a full run engine execution? i have used the stub wrapper preproccesor before when running the optimization at CHX, and it is very similar to the reject_control_messages with some extra metadata tagging but have gotten the runengine upset somewhere between the run bundler and detector staging. There was some type of sequencing issue with declaring a stream, staging/unstaging a detector, and redeclaring the same stream within the run.

thopkins32:
@MTakahashi-KWH I added an integration test to prove it works with Tiled and a simple evaluation function w/ callback.

MTakahashi-KWH (approved):
An full integration test is great.  My primary concern was if you had been able to run a testbed implementation so that any hijinks of run bundling could be caught. 

Should we tag this as an experimental utility as we move to full testing and can consider what API components actually fit the necesary abstraction?

thopkins32:
Marked as draft while I add that this as experimental.

PR #355, opened 2026-08-25
fix(plans): register evaluation failures

### LINKED ISSUE #354: `blop.plans.optimize_step` should register failures during evaluation
Currently, `blop.plans.optimize_step` only registers failures during the acquisition plan. However, this should also include failures during evaluation. You may have been able to acquire the data but not make sense of it during analysis (e.g. you acquired complete detector noise).

It's often more convenient to raise within the evaluation than try to come up with a "bad" outcome.

This is as trivial as moving the evaluation function call into the try-except block.

terminalchai:
## Summary

- register suggested points as failures when evaluation raises
- preserve the original evaluation exception and skip ingestion
- add regression coverage for evaluation failures

## Testing

- 269 non-integration tests passed
- Ruff formatting and lint passed
- Pyright passed
- All pre-commit hooks passed

Fixes #354

### DIFF
--- a/src/blop/plans.py
+++ b/src/blop/plans.py
@@ -145,12 +145,12 @@ def optimize_step(
         )
     try:
         uid = yield from acquisition_plan(suggestions, actuators, optimization_problem.sensors, *args, **kwargs)
+        outcomes = optimization_problem.evaluation_function(uid, suggestions)
     except Exception:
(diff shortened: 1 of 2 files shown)


thopkins32 (approved):
Thanks bot!

PR #356, opened 2026-08-25
Prevent route suggestions from revisiting a suggestion twice

thopkins32:
It was creating cycles in edge cases causing duplicate visits to the same suggestion.

Discussed with @MTakahashi-KWH in-person.


MTakahashi-KWH (approved):
fix to hacky initial tsp implementation by changing to algorithm that both allows selection of starting point (done by prev hacky weighting) and ensures Hamiltonian paths (backtracking paths possible with prev solvers due to their induction to complete graphs)

PR #363, opened 2026-08-31
Promote unicode numpy dtypes to at most 36 characters long

### LINKED ISSUE #362: BUG: InferredReadable inferring too strict numpy dtype for string values
When passing a string value in for the initial value of an `InferredReadable`, it infers the dtype strictly.

E.g.
```python
import numpy as np

arr = np.array(["0"])
assert arr.dtype.str == "<U1"
```
which then gets passed to `dtype_numpy` in the `InferredReadable.describe()` result. This truncates every subsequent string to be at most the same length as the initial value. Often the initial value is `"0"` (suggestion id), so this corrupts any trials with more than one digit.

thopkins32:
Assisted-by: oh-my-pi:gpt-5.6-sol

Closes #362

### DIFF
--- a/src/blop/tests/test_utils.py
+++ b/src/blop/tests/test_utils.py
@@ -22,6 +22,7 @@ def test_inferred_readable_scalar_string():
     r = InferredReadable("ids", Source.OTHER, ["0"])
     assert r.read()["ids"]["value"] == "0"
     assert r.describe()["ids"]["dtype"] == "string"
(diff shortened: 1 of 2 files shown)


MTakahashi-KWH (approved):
I have been trying to recreate this bug and cant get local tiled writers or inferred variables to reproduce this error. As this is a 0 cost fix, id suggest merging it to see if it resolves future occurrences. If not, we would be more informed of the possible source

PR #365, opened 2026-08-31
Add 'blop_acquisition_order' metadata for suggestion reordering

### LINKED ISSUE #352: Change `'blop_suggestions"` key to `"blop_acquisition_order"`
It's a bit awkward to write an evaluation function where the signature contains `suggestions` but in a possibly different order than what was actually acquired. This is a hidden detail that may be forgotten easily.

Instead, we should store something like:
 ```python
md = {
    "blop_suggestions": suggestions,              # backwards-compatible, maybe deprecated later
    "blop_acquisition_order": [s["_id"] for s in suggestions],
}
 ```

to make it clearer that acquisition order may be different than suggestion order. Then users can write evaluation functions that pull the `"blop_acquisition_order"` rather than pulling `"blop_suggestions"` and completely ignoring the `suggestions` argument every time.

As an alternative, I briefly considered changing `AcquisitionPlan`'s return type from `Hashable` -> `tuple[Hashable, Sequence[Mapping]]`, such that you must return the suggestions in the order they were evaluated, but decided against it. It assumes too much at the protocol level. For all we know, suggestions may be able to be evaluated in parallel and no order exists!

thopkins32:
Assisted-by: oh-my-pi:gpt-5.5

To be used in favor of `blop_suggestions` in evaluation functions. The intent is clearer.

Closes #352

### DIFF
--- a/src/blop/tests/integration/test_simple_experiment.py
+++ b/src/blop/tests/integration/test_simple_experiment.py
@@ -25,19 +25,15 @@ class HimmelblauEvaluation:
         self.uids.append(uid)
         run = self.tiled_client[uid]
         acquisition_order = run.start["blop_acquisition_order"]
-        suggestions_by_id = {suggestion["_id"]: suggestion for suggestion in suggestions}
         x1_data = run["primary/x1"].read()
         x2_data = run["primary/x2"].read()
 
-        assert set(acquisition_order) == set(suggestions_by_id)
-
         outcomes = []
         for index, suggestion_id in enumerate(acquisition_order):
-            suggestion = suggestions_by_id[suggestion_id]
             x1 = x1_data[index]
             x2 = x2_data[index]
             value = (x1**2 + x2 - 11) ** 2 + (x1 + x2**2 - 7) ** 2
-            outcomes.append({"_id": suggestion["_id"], "himmelblau_2d": value})
+            outcomes.append({"_id": suggestion_id, "himmelblau_2d": value})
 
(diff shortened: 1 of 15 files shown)


src/blop/tests/integration/test_simple_experiment.py
```diff
@@ -24,20 +24,16 @@ def __init__(self, tiled_client: Container) -> None:
     def __call__(self, uid: str, suggestions: list[dict]) -> list[dict]:
         self.uids.append(uid)
         run = self.tiled_client[uid]
-        reordered_suggestions = run.start["blop_suggestions"]
+        acquisition_order = run.start["blop_acquisition_order"]
         x1_data = run["primary/x1"].read()
         x2_data = run["primary/x2"].read()
 
-        assert {suggestion["_id"] for suggestion in reordered_suggestions} == {
```
MTakahashi-KWH, line 32: why was this removed? id think this was a safeguard s.t. routing will always return all supplied points. id think asserting it by rewriting it as:
{sugg["_id"] for sugg in suggestions} == set(reordered_suggestions)

is good enough

thopkins32: Ah you are right. I will put the set comparison back.

MTakahashi-KWH (approved):
out side of putting like 20 more of:
 "Suggestions are optional analysis context; do not assume their sequence matches acquired data or preserves optimizer generation order." 

i think this is good

PR #370, opened 2026-09-08
Make EvauationFunction uid argument a generic type

### LINKED ISSUE #369: Make `EvaluationFunction` generic over the acquisition reference type instead of requiring `Hashable`
### Problem

 `EvaluationFunction` was recently relaxed from accepting only a Bluesky run UID to accepting any `Hashable` acquisition identifier:

 ```python
class EvaluationFunction(Protocol):
    def __call__(
        self,
        uid: Hashable,
        suggestions: Sequence[Mapping],
    ) -> Sequence[Mapping]:
        ...
 ```

 This is a good move away from assuming every evaluation is keyed by a single Bluesky run UID.

 However, `Hashable` is still not quite the right abstraction. It describes an implementation property — usable as a dictionary key — rather than
 the real contract:

> The evaluation function receives an acquisition reference that contains enough information to find/read/evaluate the acquired data.

 Some acquisition references naturally are hashable:

 ```python
uid: str
uid: uuid.UUID
uid: tuple[str, ...]
 ```

 But other useful acquisition references may not be hashable, or may only be hashable by accident:

 ```python
@dataclass
class QueueserverAcquisition:
    correlation_uid: str
    item_uid: str | None
    plan_name: str
    metadata: Mapping[str, Any]
 ```

 For Queue Server/Tiled workflows, an evaluator may need a richer object containing correlation metadata, Queue Server item UID, plan name,
 multiple run UIDs, stream names, or backend-specific context. Forcing that object to be hashable adds an unnecessary constraint if Blop itself
 does not need to use it as a dictionary key.

 There is also a static typing issue. This protocol requires implementations to accept any `Hashable`, but many valid evaluators only accept a
 narrower acquisition reference type:

 ```python
class RunUidEvaluator:
    def __call__(
        self,
        uid: str,
        suggestions: Sequence[Mapping],
    ) -> Sequence[Mapping]:
        ...
 ```

 That evaluator is valid conceptually, but it may not satisfy `EvaluationFunction` under type checkers because `str` is narrower than `Hashable`.

 ### Proposal

 Make `EvaluationFunction` generic over the acquisition reference type.

 Sketch:

 ```python
from collections.abc import Mapping, Sequence
from typing import Protocol, TypeVar

TAcquisition = TypeVar("TAcquisition", contravariant=True)


class EvaluationFunction(Protocol[TAcquisition]):
    def __call__(
        self,
        acquisition: TAcquisition,
        suggestions: Sequence[Mapping],
    ) -> Sequence[Mapping]:
        ...
 ```

 Then specific evaluators can type their true input:

 ```python
class RunUidEvaluator:
    def __call__(
        self,
        acquisition: str,
        suggestions: Sequence[Mapping],
    ) -> Sequence[Mapping]:
        ...
 ```

 ```python
@dataclass(frozen=True, slots=True)
class QueueserverAcquisition:
    correlation_uid: str
    item_uid: str | None
    plan_name: str


class TiledQueueserverEvaluator:
    def __call__(
        self,
        acquisition: QueueserverAcquisition,
        suggestions: Sequence[Mapping],
    ) -> Sequence[Mapping]:
        ...
 ```

 The acquisition plan protocol can be generic over the same reference type:

 ```python
from typing import TypeVar

TAcquisition_co = TypeVar("TAcquisition_co", covariant=True)


class AcquisitionPlan(Protocol[TAcquisition_co]):
    def __call__(...) -> MsgGenerator[TAcquisition_co]:
        ...
 ```

 And the optimization problem ties the two together:

 ```python
@dataclass(frozen=True)
class OptimizationProblem(Generic[TAcquisition]):
    acquisition_plan: AcquisitionPlan[TAcquisition]
    evaluation_function: EvaluationFunction[TAcquisition]
 ```

 For Queue Server, the acquisition reference type could be independent of the Bluesky plan return type because Queue Server submission itself
 produces useful identity:

 ```python
@dataclass(frozen=True, slots=True)
class QueueserverAcquisition:
    correlation_uid: str
    item_uid: str | None
    plan_name: str
 ```

 ```python
evaluation_function: EvaluationFunction[QueueserverAcquisition]
 ```

 ### Naming

 Consider renaming the first parameter from `uid` to `acquisition` or `acquisition_id`.

 `uid` still suggests a single Bluesky run UID, while the generalized protocol supports:

 - a Bluesky run UID,
 - a tuple of event UIDs,
 - a Queue Server acquisition token,
 - a Tiled correlation key,
 - a backend-specific acquisition reference object.

 Possible protocol signature:

 ```python
class EvaluationFunction(Protocol[TAcquisition]):
    def __call__(
        self,
        acquisition: TAcquisition,
        suggestions: Sequence[Mapping],
    ) -> Sequence[Mapping]:
        ...
 ```

 ### Compatibility

 A generic type can preserve the current behavior with a default type parameter if the supported Python/type-checking baseline allows it:

 ```python
class EvaluationFunction(Protocol[TAcquisition]):
    ...
 ```

 with aliases such as:

 ```python
RunUidEvaluationFunction = EvaluationFunction[str]
HashableEvaluationFunction = EvaluationFunction[Hashable]
 ```

 If defaulted type parameters are not desirable yet, existing user code can continue to work structurally. Most runtime users will not need to
 change anything because protocols are structural and parameter names are not part of positional call compatibility.

 ### Benefits

 - Removes an unnecessary hashability constraint.
 - Accurately models the contract: evaluator-defined acquisition reference in, outcomes out.
 - Supports richer Queue Server/Tiled acquisition tokens.
 - Improves static typing for evaluators that only accept `str`, `uuid.UUID`, or a specific `dataclass`.
 - Aligns `AcquisitionPlan` and `EvaluationFunction` around one acquisition reference type.
 - Avoids continuing the implicit “single Bluesky run UID” assumption through the name `uid`.

 ### Acceptance criteria

 - `EvaluationFunction` is generic over the acquisition reference type.
 - `AcquisitionPlan` and `OptimizationProblem` connect acquisition return type to evaluator input type where applicable.
 - Existing run-UID evaluators remain valid.
 - Evaluators accepting a richer acquisition token can be typed without requiring the token to be `Hashable`.
 - Documentation describes the first argument as an acquisition reference, not necessarily a Bluesky run UID.

thopkins32:
Assisted-by: oh-my-pi:gpt-5.5

Closes #369 

See the linked issue for the problem and why this PR was needed.

~~In addition to the issue above, I've also enabled `dict` to be inferred for `acquisition_uid` in the `read_step` plan stub. This allows you to store JSON serializable dictionaries per-event.~~

^^This wasn't a good idea and it doesn't really fit well with event-model.

### DIFF
--- a/docs/source/explanations/acquisition-plan.rst
+++ b/docs/source/explanations/acquisition-plan.rst
@@ -25,18 +25,18 @@ association between optimizer suggestions and the acquired data within the
 chosen storage or event system (typically stashing the suggestions in proper 
 order in storage or tagging the run markdown).
 
-The plan must return a hashable acquisition identifier. Blop passes that value
-unchanged to the evaluation function. A Bluesky run UID is the usual identifier
-for run-owning acquisitions. The in-run default returns suggestion IDs in the
-order they were executed, but custom plans may return any hashable key their
-matching evaluation function understands.
+The plan must return a uid. Blop passes that value unchanged to the evaluation
+function. A Bluesky run UID is the usual value for run-owning acquisitions. The
+in-run default returns suggestion IDs in the order they were executed, but
+custom plans may return any concrete uid type their matching evaluation
+function understands.
 
 A simple example that optimizes in a subspace while executing measurements in
 the full physical coordinate system is shown below. 
 
 .. code-block:: python
 
-    from collections.abc import Hashable, Mapping, Sequence
+    from collections.abc import Mapping, Sequence
     from typing import Any
 
     from bluesky.utils import MsgGenerator
@@ -49,7 +49,7 @@ the full physical coordinate system is shown below.
             actuators: Sequence[Actuator],
             sensors: Sequence[Sensor] | None = None,
             md: Mapping[str, Any] | None = None,
-        ) -> MsgGenerator[Hashable]:
+        ) -> MsgGenerator[str]:
 
             coords = [subspace_to_real(s) for s in suggestions]
             return (yield from default_acquire(
--- a/docs/source/explanations/callbacks.rst
+++ b/docs/source/explanations/callbacks.rst
@@ -54,9 +54,9 @@ but the contents are optimization-focused.
     values, and outcome values. For batched optimization, these values may be
     arrays.
 
-For ``acquisition_uid``, native array-like identifiers such as a tuple of event
-UIDs are stored directly. Other hashable identifiers are stored using their
-``repr``; the evaluation function still receives the original identifier.
+For ``acquisition_uid``, native array-like UIDs are stored directly. Mapping and
+dataclass UIDs that serialize to JSON are stored as dictionaries. Other UIDs are
+stored using their ``repr``; the evaluation function still receives the original value.
 
 ``stop``
     Completion status and reason, useful for summaries or cleanup.
--- a/docs/source/explanations/callbacks.rst
+++ b/docs/source/explanations/callbacks.rst
@@ -54,8 +54,7 @@ but the contents are optimization-focused.
     values, and outcome values. For batched optimization, these values may be
     arrays.
 
-For ``acquisition_uid``, native array-like UIDs are stored directly. Mapping and
-dataclass UIDs that serialize to JSON are stored as JSON strings. Other UIDs are
+For ``acquisition_uid``, native array-like UIDs are stored directly. Other UIDs are
 stored using their ``repr``; the evaluation function still receives the original value.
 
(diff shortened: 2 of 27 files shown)


MTakahashi-KWH (approved):
Looks fine. Beneficial that acquisition plans once again have more assertion of uid parameter types during development (instead of the runtime validation of uid within the previous workflow and tutorials)

side note: explanations/acquisition-plan probably needs an update in its descriptions as some context has been lost over various edits.

PR #374, opened 2026-09-10
adding explicit scipy dependency

### LINKED ISSUE #373: Scipy missing from dependencies
We added Scipy.optimize support a while ago but it wasn't noticed that Scipy is not listed as a dependency as Ax carries Scipy into the package for its own use. Its best to make it an explicit optional dependency even if its just:

`scipy = ["scipy==*"]`

in the pyproject file.

MTakahashi-KWH:
added scipy as an optional dependency so that the blop[ax] component isn't a hidden dependency of blop.scipy

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -90,6 +90,7 @@ nb_render_image_options = {
     "align": "center",
 }
 nb_execution_mode = "auto"
+nb_execution_show_tb = True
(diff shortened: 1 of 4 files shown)


MTakahashi-KWH:
targeting #373

thopkins32 (approved):

PR #376, opened 2026-09-11
Flatten read step

### LINKED ISSUE #375: flatten Read_step so that each trial is a row
# Problem

Currently, each batch iteration in Read_step is recorded as a single row in the event model (and in the resulting Tiled DataFrame).

To normalize the number of suggestions per row, tiled_writer / event modeling currently relies on null/empty packing of the suggestion_ids array. This approach has proven fragile and makes downstream analysis more cumbersome than necessary.

# Proposed Change

Split each of the trials in an iteration into independent rows, and subsequently into individual trigger and read calls. That removes the need for any padding.

The loss of explicit per-iteration granularity should be relatively low impact, since the iterations can still be reconstructed/grouped using uid when needed.

MTakahashi-KWH:
closes #375 also resolving incompatible behavior with upstream libraries

### DIFF
--- a/src/blop/plan_stubs.py
+++ b/src/blop/plan_stubs.py
@@ -1,7 +1,6 @@
 """Bluesky plan stubs for optimization."""
 
 import logging
-from collections import defaultdict
 from collections.abc import Hashable, Mapping, MutableMapping, Sequence
 from typing import Any, Literal, cast
 
@@ -97,7 +96,9 @@ def read_step(
     stream_name : str, optional
         Event stream name for the optimization tracking event.
     """
-    # Group by ID_KEY to get proper suggestion/outcome order
+    # Normalize the acquisition UID for event-model storage.
+    normalized_uid = _acquisition_identifier_value(uid)
+
     suggestion_by_id = {}
     outcome_by_id = {}
     for suggestion in suggestions:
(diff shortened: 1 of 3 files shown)


src/blop/tests/test_plans.py
```diff
@@ -359,31 +359,22 @@ def test_optimize_multiple_with_n_points(RE):
...
-            list(data["objective"])
-            if hasattr(data["objective"], "__iter__") and not isinstance(data["objective"], str)
-            else [data["objective"]]
-        )
-        assert x1_vals == [0.0, 0.1]
-        assert obj_vals == [0.0, 0.1]
+
+        assert int(data["suggestion_ids"]) in {0, 1}
+        assert data["x1"] in {0.0, 0.1}
+        assert data["objective"] in {0.0, 0.1}
```
thopkins32, line 372: Can you make these assertions more explicit? Set membership is not the contract under test, order matters here.

src/blop/tests/test_plans.py
```diff
@@ -436,12 +427,12 @@ def _to_list(x):
...
-        assert _to_list(data["x3"]) == [0.0, 0.3]
-        assert _to_list(data["objective1"]) == [0.0, 0.1]
-        assert _to_list(data["objective2"]) == [0.1, 0.2]
-        assert _to_list(data["suggestion_ids"]) == ["0", "1"]
+        assert data["x1"] in [0.0, 0.1]
+        assert data["x2"] in [0.0, 0.2]
+        assert data["x3"] in [0.0, 0.3]
+        assert data["objective1"] in [0.0, 0.1]
+        assert data["objective2"] in [0.1, 0.2]
+        assert data["suggestion_ids"] in ["0", "1"]
```
thopkins32, line 435: Same with these.

thopkins32 (changes requested):
The integration test is failing on an assertion. Run with `pixi run integration-tests`.

You will also have to update the `OptimizationLogger` as it is directly linked with `read_step`. It's slightly concerning that the logger tests didn't pick this up...

The docs still mention the array/padding behavior. See `src/blop/plan_stubs.py:78-93`, `docs/source/explanations/callbacks.rst:52-55`, and `src/blop/callbacks/logger.py:77-79`.

src/blop/callbacks/logger.py
```diff
@@ -172,114 +187,83 @@ def event(self, doc: Event) -> Event:
...
-                summary.append("  min: ", style=_DIM_STYLE)
-                summary.append(_format_stat(s.min))
-                summary.append("  max: ", style=_DIM_STYLE)
-                summary.append(_format_stat(s.max))
-                summary.append("  mean: ", style=_DIM_STYLE)
-                summary.append(_format_stat(s.mean))
-            summary.append(f"\n  ({outcome_point_count} pts sampled)", style=_DIM_STYLE)
-            self._console.print(summary)
-
+        if self._current_step % 5 == 4:
```
thopkins32, line 240: Why would you do it this way? `self._current_step` is incremented before this so summaries are going to be printed at steps 4, 9, 14, ... rather than 5, 10, 15.
```suggestion
        if self._current_step % 5 == 0:
```

src/blop/callbacks/logger.py
```diff
@@ -172,114 +187,83 @@ def event(self, doc: Event) -> Event:
...
-        )
-        # Filter out NaN-padded entries: suggestion_ids padded with "" indicate padding
-        if suggestion_ids:
-            valid_indices = [i for i, sid in enumerate(suggestion_ids) if sid != "" and str(sid).strip() != ""]
-        else:
-            valid_indices = list(range(n_total))
-        n_valid = len(valid_indices) if valid_indices else n_total
+        if isinstance(acquire_uid, Hashable) and acquire_uid not in self._seen_uids:
+            self._seen_uids.add(acquire_uid)
+            self._current_iteration += 1
```
thopkins32, line 206: We have to figure out something better for acquisition UID or simply exclude it from the callback altogether.

Also, if the `acquire_uid` is a list we also only use the first element of the list, which is incorrect.

Maybe `read_step` should be producing an `iteration` data key in the event document. We can use that for iteration tracking instead of checking for unique acquisition UIDs.

src/blop/tests/callbacks/test_logger.py
```diff
@@ -134,7 +134,7 @@ def test_event_scalar_data(logger, console):
     result = logger.event(doc)
     assert result is doc
     assert console.print.call_count >= 1
-    console.rule.assert_called_once()
+    # removed rule assertion as rulered values are now based on number of points rather than per iteration
```
thopkins32, line 137: Why keep a comment here?
```suggestion
```

src/blop/tests/callbacks/test_logger.py
```diff
@@ -144,7 +144,7 @@ def test_event_without_iteration_limit_omits_total(logger, console):
 
     logger.event(_make_event(data={"x": 1.5, "y": 3.14}))
 
-    console.rule.assert_called_once_with("Iteration 1", style="blue")
+    # removed rule assertion as rulered values are now based on number of points rather than per iteration
```
thopkins32, line 147:
```suggestion
```

src/blop/tests/callbacks/test_logger.py
```diff
@@ -179,8 +179,9 @@ def test_event_batch_with_nan_padding(logger, console):
 def test_event_multiple_iterations(logger, console):
     """Successive events should accumulate without error."""
     _setup_descriptor(logger)
-    logger.event(_make_event(data={"x": 1.0, "y": 10.0}))
-    logger.event(_make_event(data={"x": 3.0, "y": 20.0}))
+    for _ in range(10):
+        logger.event(_make_event(data={"x": 1.0, "y": 10.0}))
+        logger.event(_make_event(data={"x": 3.0, "y": 20.0}))
     assert console.print.call_count >= 1
```
thopkins32, line 185: This isn't testing multiple iterations and merely asserting the the `call_count >= 1` is doing nothing.

src/blop/plan_stubs.py
thopkins32, line 92: Remove parameter `n_points` which is no longer used.

thopkins32 (changes requested):

src/blop/tests/callbacks/test_logger.py
MTakahashi-KWH: This is a codecov issue where for some reason previous reports said that the code run every 5 steps was unreached. At least 10 iters was required to cover this, even though the test assertion of rule was called at least twice (top and bottom rule) was passing.

src/blop/callbacks/logger.py
MTakahashi-KWH: i was teetering on not adding another iteration inferred variable, given that we could group by uid repr. But, yes, its better explicit

src/blop/callbacks/logger.py
MTakahashi-KWH: Had a mental edge case thought where it would immediately print 0 count stats but it actually preincrements, so ill fix.

PR #378, opened 2026-09-15
Add Bluesky checkpoints per iteration

### LINKED ISSUE #367: Use of `bluesky.plan_stubs.checkpoint` in plans
It is often the case that a user may want to pause an optimization run to inspect optimizer state, compute analyses, etc. before deciding to `RE.resume()` or `RE.stop()`. Bluesky's run engine allows you to defer a pause by hitting Ctrl + C once.

It would be nice to put a `bluesky.plan_stubs.checkpoint` call after each iteration (and possibly strip checkpoints from `list_scan` in `default_acquire`) so that when a user requests a pause, it occurs at the end of an iteration.

Personally, I wish there was a way to distinguish between different checkpoints. One may want to pause a Bluesky run for many different reasons, and the next checkpoint may not always be the desired stopping point.

josephhanrahan:
This PR uses the `msg_mutator` preprocessor to remove Bluesky checkpoint messages from being yielded in `default_acquire` instead yielding them per optimization iteration in `optimize`. This allows optimizations to pause at meaningful iteration boundaries instead of at an arbitrary point within an iteration.

Closes #367.


src/blop/tests/test_plans.py
```diff
@@ -1319,3 +1330,18 @@ def should_stop(self) -> tuple[bool, str | None]:
...
+    optimization_problem = OptimizationProblem(
+        MagicMock(spec=Optimizer),
+        actuators=[MovableSignal("x1")],
+        sensors=[ReadableSignal("objective")],
+        evaluation_function=MagicMock(spec=EvaluationFunction),
+    )
+    commands = []
+
+    with patch.object(RE, "msg_hook", lambda msg: commands.append(msg.command)):
+        RE(optimize(optimization_problem, iterations=5))
```
thopkins32: Simplification, no need for patching.
```suggestion
    RE.msg_hook = lambda msg: commands.append(msg.command)
    RE(optimize(optimization_problem, iterations=5))
```

src/blop/tests/test_plans.py
```diff
@@ -1124,6 +1124,17 @@ def test_default_acquire_multiple_movables_readables(RE):
     assert movable2.read()["x2"]["value"] == 0.1
 
 
+def test_default_acquire_checkpoint_removal(RE):
+    movable = MovableSignal("x1", initial_value=-1.0)
+    readable = ReadableSignal("objective")
+    commands = []
+
+    with patch.object(RE, "msg_hook", lambda msg: commands.append(msg.command)):
+        RE(default_acquire([{"x1": 0.0, "_id": 0}, {"x1": 0.1, "_id": 1}], [movable], [readable]))
```
thopkins32:
```suggestion
    RE.msg_hook = lambda msg: commands.append(msg.command)
    RE(default_acquire([{"x1": 0.0, "_id": 0}, {"x1": 0.1, "_id": 1}], [movable], [readable]))
```

thopkins32 (changes requested):

thopkins32 (approved):
docs build failure was pre-existing

